PR #181, opened 2022-12-08
Resolves #gh-173 MEQL tests

NigelPalmer:


src/app/pipes/meql.pipe.spec.ts
```diff
@@ -0,0 +1,356 @@
...
+    const line7 = `${tab})${newline}`;
+    const line8 = ')';
+    return [
+      pipe.transform(query),
+      `${line1}${line2}${line3}${line4}${line5}${line6}${line7}${line8}`,
+    ];
+  }
+}
+
+describe('JqlPipe', () => {
```
pjmonks, line 94: Changing the name of the pipe meant this was not updated, please correct

src/app/pipes/meql.pipe.spec.ts
```diff
@@ -0,0 +1,356 @@
...
+  });
+
+  it('transforms field equal to string', () => {
+    const testResult = meqlTestHelper.singleConditionTest(
+      pipe,
+      'Simple String Field',
+      '=',
+      'Some Text'
+    );
+    expect(testResult[0]).toBe(testResult[1]);
```
pjmonks, line 114: No problem here, but there is another way to do this which is hopefully more self-explanatory. Looking at your test helper, the function returns an array with just 2 objects in it, so it is effectively a tuple. You can destructure tuples in JS/TS into individual variables, so you could write this test like so:

```ts
it('transforms field equal to string', () => {
    const [actual, expected] = meqlTestHelper.singleConditionTest(
      pipe,
      'Simple String Field',
      '=',
      'Some Text'
    );
    expect(actual).toBe(expected);
  });
```

The reason I suggest it is because referring to `actual` and `expected` seems clearer than understanding what `testResult[0]` and `testResult[1]` is.

src/app/pipes/meql.pipe.spec.ts
```diff
@@ -0,0 +1,356 @@
...
+    const testResult = meqlTestHelper.singleConditionTest(
+      pipe,
+      'Simple String Field',
+      '!=',
+      'Some Text'
+    );
+    expect(testResult[0]).toBe(testResult[1]);
+  });
+
+  it('transforms field equal to number', () => {
```
pjmonks, line 127: If you're familiar with data-driven tests, then there are ways in Jest to write those and parameterise some of your tests:

```ts
// Run this test on each value passed to each()
it.each([1, 2, 3])('should print %p'), (value) => {
  // Use the value here
});
```

I was wondering if some of the repetition I've seen could be applied here.

src/app/pipes/meql.pipe.spec.ts
```diff
@@ -0,0 +1,356 @@
...
+    const testResult = meqlTestHelper.singleConditionTest(
+      pipe,
+      'Simple String Field',
+      '!=',
+      99
+    );
+    expect(testResult[0]).toBe(testResult[1]);
+  });
+
+  it('transforms and uses "and" condition', () => {
```
pjmonks, line 147: Another Jest feature is you can nest `describe()` calls. Here I see you have lots of tests but maybe they could be grouped together:

```ts
describe('MeqlPipe', () => {
  describe('single condition tests', () => {
    // it() calls go here...
  });

  describe('AND condition tests', () => {
    // it() calls go here...
  });

  // etc...
});
```

Not necessary, depends on how you want to structure your tests.

pjmonks (approved):
All looks great 👍 My feedback comments are only suggestions since you're new to Jest, but it all looks fine regardless and passes tests, which is more important.

Approved ✔️

src/app/pipes/meql.pipe.spec.ts
NigelPalmer: Corrected

src/app/pipes/meql.pipe.spec.ts
NigelPalmer: I agree this would be clearer. Updated as suggested.

src/app/pipes/meql.pipe.spec.ts
NigelPalmer: Test groups created. It looks neater and is easier to follow.

src/app/pipes/meql.pipe.spec.ts
NigelPalmer: ...yet superseded by further refactoring so no longer relevant.

pjmonks (approved):
Approved ✔️

Issue #168, opened 2022-12-05
Redesign the explorer UI to include the Query Builder in a sensible way.


NigelPalmer:
# Background

The current design of the "My Requests" page lists each request name on the left, with the detail view of the selected request on the right (majority) of the page. When introducing the cohort builder though, this will take up more screen estate and make the already cramped looking screen potentially worse.

# List/Detail View

Instead, the page should be redesigned into a more typical list/detail view - similar to how account pages would appear on other websites.

- The "My accounts" page will lead to the "list" view, displaying in a table:
  - Each request name
  - Status
  - Ancillary details? e.g. dates? (Not essential)
  - Typical CRUD controls e.g. "View", "Edit", Delete" etc
- Selecting a single view will then redirect to a "detail" view of that request, showing the same information as before:
  - List of data points chosen
  - Controls to manage those data points
  - Description
  - Submit actions
  - etc

# Wireframes

![image](https://user-images.githubusercontent.com/85611988/205659193-31a57538-c66d-4b24-b8f8-dd1a356a12db.png)
![image](https://user-images.githubusercontent.com/85611988/205659238-10d29e01-6298-475b-b06f-82cc61bd0c7b.png)

# Routing

This will mean a change to the routing as well:

`/requests` - directs to the list view of all requests
`/requests/{uuid}` - directs to the detail view of a request, defined by it's `{uuid}`

# Out of scope

Other than rearranging the pages, the existing functionality should remain. No new functionality should be added as part of this particular issue.

PR #183, opened 2022-12-19
gh-164 Create and update rules to save data request queries

### LINKED ISSUE #164: Persist query in mauro database.
The data behind the query builder UI can be simply piped into a json format. Json is the most sensible version of the data to save, since it gives us the most flexibility in transforming the data into the three different desired formats, UI representation, Simple Query Language, and SQL. Two queries will be required to be saved, Cohort and Data. Two new fields need to be added to the datamodel.data_model table to hold the json data for these queries. Then api requests will need to be identified and updated to allow the new fields to be inserted and updated as required. Api requests that return data from the datamodel.data_model table will also need updating to include the new fields.

pjmonks:
Resolves #164 

- Create RulesService
- Add service functions to get single Data Request
- Move MeqlPipe to DataExplorerModule
- Create mdm-meql-output component
- New page to display individual query by type for a data request
- Update the Query Builder
  - Use mdm-meql-output component
  - Remove title
  - Remove show/hide button
  - Remove unecessary CSS
  - Make colour palette an input option to the component
  - Change how input of data elements affects initialisation of component
  - Simplify sorting of data elements
  - Add two-way binding to query builder
- Save queries to the data request
- Make sure queries are loaded from data requests correctly
- Add guard to prevent navigation if changes were not saved

To test at the moment, manually go to the URL http://localhost:4201/#/requests/{requestId}/queries/{queryType}, where:

- `requestId` is the UUID of the request data model
- `queryType` is either `cohort` or `data`

Still to do:

1. Combine with other pages better when they are ready, introduce consistency in UI
2. Add correct back link to the request detail page which is currently missing, being built in #168 

# Screenshots

image


src/app/pages/request-query/request-query.component.ts
```diff
@@ -0,0 +1,148 @@
...
+      .pipe(
+        catchError(() => {
+          this.toastr.error('There was a problem saving your query.');
+          return EMPTY;
+        }),
+        finalize(() => this.broadcast.loading({ isLoading: false }))
+      )
+      .subscribe((submitted) => {
+        this.toastr.success(`Your ${this.queryType} query was saved successfully.`);
+        // TODO: change below to a redirect
```
NigelPalmer, line 137: If the query is saved successfully, I don't think a redirect should then take place. A user may wish to save as they go, not be instantly moved elsewhere as soon as they click the save button.

NigelPalmer (commented):
- [X] Create RulesService
- [X] Add service functions to get single Data Request
- [X] Move MeqlPipe to DataExplorerModule
- [X] Create mdm-meql-output component
- [X] New page to display individual query by type for a data request
- [X] Update the Query Builder
  - [X] Use mdm-meql-output component
  - [X] Remove title
  - [X] Remove show/hide button
  - [X] Remove unecessary CSS
  - [X] Make colour palette an input option to the component
  - [X] Change how input of data elements affects initialisation of component
  - [X] Simplify sorting of data elements
  - [X] Add two-way binding to query builder
- [X] Save queries to the data request
- [X] Make sure queries are loaded from data requests correctly
- [ ] Add guard to prevent navigation if changes were not saved. 
_**Issue**: This works, but after clicking "Save changes" the "modified" state remains so the user is always informed that they have unsaved changes when they do not. Looking at the code this may be intentional since there appears to be an assumption that once these changes are merged in with the new page workflows, the user will be redirected elsewhere after clicking save. But, I don't think this should be the case. See my notes in the code for further information. **Please fix**._

src/app/pages/request-query/request-query.component.ts
pjmonks: Changed this and the tests to reflect what you've asked for.

NigelPalmer: This all looks good now. Ready for merge.

NigelPalmer (approved):

PR #184, opened 2022-12-20
gh-170 Adjustments to query builder controls

### LINKED ISSUE #170: Query Builder UI controls should autosize to fit content
The query builder uses 3 controls for each line of the query. These are:
<field name> <operator> <value>

At the moment these controls have a fixed width. They should resize when the query definition is being loaded so that the data inside them does not get truncated.

pjmonks:
- Change widths of query field controls to use percentages
- Change query builder boolean option to a select control

Resolves #170 

Can only set percentages for control widths because the query builder library component controls the container which all controls are in, there is no template available to change that. Also, the Angular Material mat-select is not able to have an autosized width based on content in the option list.

image

### DIFF
--- a/src/app/data-explorer/querybuilder/querybuilder.component.html
+++ b/src/app/data-explorer/querybuilder/querybuilder.component.html
@@ -129,11 +129,15 @@ SPDX-License-Identifier: Apache-2.0
       </ng-container>
       <!-- Boolean -->
       <ng-container *queryInput="let rule; type: 'boolean'; let onChange = onChange">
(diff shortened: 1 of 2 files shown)


NigelPalmer (approved):
- [x] Controls resize more sensibly now as the browser window changes size. I can see some ways that we could make the user experience smoother, regarding the resizing of fields, but it is not worth pursuing this until we know how users actually use the site.
- [x] Booleans control is now a drop down and this improves the UX.
- [x] All code changes look sensible to me.
**Passed**

PR #185, opened 2022-12-20
gh-156 Synchronise unsent request count for header

### LINKED ISSUE #156: Shopping Basket shows wrong number of requests
The shopping basket increments the number of items in the basket when an existing request has new fields added to it. This results in a mismatch between the real number of requests and the total being shown by the shopping trolly. In this example there are only 4 requests.
image

pjmonks:
When `data-request-added` or `data-request-submitted` broadcast events happen, they now directly check the request count from the server.

Fixes issue where the wrong count would happen because `data-request-added` event was used to alert a refresh of other data, caused a side effect of breaking the header counter.

Fixes #156


NigelPalmer (approved):
- [x] Request count number does not increase when new fields are added.
- [x] Request count number increases when a new request is made.
- [x] Request count number decreases when a request is submitted.
- [x] Code changes look sensible to me.
**Passed**

PR #186, opened 2022-12-21
gh-155 Improve pagination control on smaller screen widths

### LINKED ISSUE #155: Search results page selector: Messy layout at smaller window sizes.
The page selector of the list results page does not resize well at smaller browser window sizes. The "With selected..." button ends up behind other buttons.
image

pjmonks:
- Reduce number of page numbers to display
- Use flexbox container to forces select and page controls to wrap when screen width is too small
- Remove total number of pages count from pagination component, seems superfluous

Fixes #155 

# Screenshots

Large screen width:

image

Small screen width:

image

### DIFF
--- a/src/app/data-explorer/data-explorer.module.ts
+++ b/src/app/data-explorer/data-explorer.module.ts
@@ -84,7 +84,7 @@ import { MeqlOutputComponent } from './meql-output/meql-output.component';
       provide: PAGINATION_CONFIG,
       useValue: {
         defaultPageSize: 10,
-        maxPagesToShow: 10,
+        maxPagesToShow: 5,
       },
     },
   ],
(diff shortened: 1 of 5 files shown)


NigelPalmer (commented):
- [X] Page resizing works without controls overlaying each other.
- [X] Pages are now limited to 5 being displayed.
- [ ] I disagree that the total page number count is superfluous. When I had 100 items, there were 10 pages, but the page indicator topped out at 5, so it was not clear that there are more than 5 pages. The way I am used to this information being presented is: 
image Could we adapt the pagination control to work in such a way?

pjmonks:
I've changed the pagination control to include the page counter again at least, updated as well to show this:

image

This is the best I can do on narrow screen widths, but the expectation is this probably won't be viewed via mobile for the foreseeable future.

NigelPalmer (approved):
- [X] Commented out code removed.
- [X] Page count restored.
**Passed**

PR #187, opened 2023-01-04
Feature/gh 168 myrequests

OButlerOcc:
Split myrequests page into a myRequests list page and a Request detail page.


pjmonks:
The new "My Requests" page now looks a lot neater for sure. There are a few design issues I see though:

1. The space between "My Requests" heading and the request list looks quite large, compared against the "My Bookmarks" page
2. It would be nicer if each request in the list had some space between them.
3. You've used a "View Details" button whereas other lists of with this style use a "View Details" hyperlink. I think I actually prefer the button because it's obviously an action - though I think I read somewhere that UX design prefers links for navigation over buttons (could be wrong here). Whichever changes, I think it would be good to make everything consistent, but as a separate issue. @NigelPalmer and @jamesrwelch - which do you prefer?

image

Compared to "My Bookmarks" page:

image

pjmonks:
image

Something has gone wrong here in the request detail page, the footer is stuck at one spot while the list continues down the page. Perhaps check the Bootstrap column/row div classes to make sure they are set correctly.

pjmonks:
Requests that are in the submitted state no longer have a "Copy" button anymore.

New:

image

Original:

image

src/app/pages/my-request-detail/my-request-detail.component.html
```diff
@@ -0,0 +1,74 @@
...
+
+SPDX-License-Identifier: Apache-2.0
+-->
+<div *ngIf="request">
+  <div class="main-row hero">
+    <h1>{{ request.label }}</h1>
+  </div>
+  <div class="container hero-top-spacer">
+    <div *ngIf="state === 'loading'" class="mdm-my-request__progress">
+      <mat-progress-bar color="primary" mode="indeterminate"></mat-progress-bar>
```
pjmonks, line 26: I don't think I've seen a progress bar (or any progress indicator) when the page is loading. Is this correct? Because it would be good to have one.

src/app/pages/my-request-detail/my-request-detail.component.ts
```diff
@@ -0,0 +1,412 @@
...
+        }),
+        catchError((error) => {
+          this.toastr.error(`Invalid Request Id. ${error}`);
+          this.state = 'idle';
+          return EMPTY;
+        })
+      )
+      .subscribe((request) => {
+        this.setRequest(request);
+        this.state = 'idle';
```
pjmonks, line 112: I'd suggest putting this state change in a `finalize()` operator in the `pipe()`. It will always be called at the end of the pipe (but before the "next" operation in the `subscribe()`).

src/app/pages/my-request-detail/my-request-detail.component.ts
```diff
@@ -0,0 +1,412 @@
...
+          return EMPTY;
+        }
+      })
+    );
+  }
+
+  private processRemoveDataElementResponse(result: boolean, message: string) {
+    if (result === false) {
+      this.toastr.error(message, 'Data element removal');
+      console.log(message);
```
pjmonks, line 195: This should be removed now if it's working and just for testing.

pjmonks (changes requested):
I think the code is pretty much there. The only issues I've found are graphical and I've added screenshots to show them to this PR in comments. 

I'd suggest anything I've raised that is critical be done as part of this review now, the rest could be siphoned off to new issues to create and revisit later - I've already highlighted one bug which was already present in the original version.

OButlerOcc:
Fixed all of these bar the filters.

Note, footer issue caused by max height in pages, might be applicable elsewhere.

ToDo:
Unsure how to submit a request to check if the copy button is working.
Cannot find filters in the previous version of the requests page, the functionality I ported across but I'm not sure what the control should look like to use it.

pjmonks (approved):
All the changes look better now 👍 

I'm going to merge and we can discuss as a team further steps to unify these changes with other work.

Issue #171, opened 2022-12-06
Query Builder field type mapping management


NigelPalmer:
The data model allows types of fields to be defined. These types could contain any text. We need to be able to map the text that is entered to a type that the query builder recognises. At the moment this has been done with a hard coded list, but this should be handled within the data model by saving the mapping between the entered type and the query builder type. Some types will not be supported by the query builder such as binary fields.

The query builder categories are:
1. category - A lookup of values. If the data end point has an enumeration list, then this option is automatically selected.
2. string - String types, e.g: char, varchar, text
3. number - Number types, e.g: decimal, float, int
4. date - Date types, e,g: date, datetime
5. time - Time types, e.g: time
6. boolean - Boolean types, e.g: boolean, bit
7. unsupported - Unsupported types, e.g: binary, cursor, table

The current hard coded list can be found in the queryBuilderDataType function.

We need to design and implement an approach to allow users who create data models to manage these mappings.

Issue #169, opened 2022-12-05
search-as-you-type required for 'model references'


NigelPalmer:
Some data elements that use the 'model reference' type will point at a terminology - usually ICD10 or OPCS4. Terminologies may have a few thousand terms, so a standard checkbox won't work for those, but we can probably make a search-as-you-type select box work for most use cases, alongside the ability to add a search term (prefix / regex / whatever). An api call would be needed to get the list of options since we wouldn't want a page to load 1000s of items. A material design search-as-you-type control should be able to be added to the query builder.

pjmonks:
@jamesrwelch As I mentioned in #166 , it would be useful if we could be provided a sample data model which represents this scenario of having a large number of terms.

PR #190, opened 2023-01-06
gh-171: Query Builder field type mapping management

NigelPalmer:
This code is reliant on the code committed in https://github.com/MauroDataMapper-Plugins/mdm-plugin-explorer/pull/13 . 

To test this a profile needs to be created on a data type to map the primitive data type to the query builder data type.

In the MDM UI
1. Go to the list of Data Types.
2. Click on one of the primitive types, e.g: float.
3. On the Description tab in the dropdown select "Add new profile..."
4. Then choose "Mauro Data Explorer - Query Builder Data Type (1.0.0)" and Save Changes
5. On the next screen that appears choose the query builder type you want to associate with the field (e.g: number) and save.

In the MDM Explorer
1. Go to a request (or create a new one) that has a float field included in it.
2. Go to the query builder. The float field will be available to be added to the query.

### DIFF
--- a/src/app/mauro/query-builder.service.ts
+++ b/src/app/mauro/query-builder.service.ts
@@ -47,7 +47,7 @@ export class QueryBuilderService {
   public setupConfig(
     dataElements: DataElementSearchResult[],
     query?: DataRequestQueryPayload
-  ): Observable<[QueryConfiguration]> {
+  ): Observable<QueryConfiguration> {
     const requests$ = dataElements.map((dataElement) => {
       const profile$ = this.getQueryBuilderDatatype(dataElement.dataType).pipe(
         catchError((error) => {
@@ -66,7 +66,7 @@ export class QueryBuilderService {
 
     return forkJoin(requests$).pipe(
       map((items) => {
-        return [this.getQueryFields(items, dataElements, query)];
+        return this.getQueryFields(items, dataElements, query);
       })
     );
   }
--- a/src/app/mauro/query-builder.service.ts
+++ b/src/app/mauro/query-builder.service.ts
@@ -1,0 +1,145 @@
+/*
+Copyright 2022 University of Oxford
+and Health and Social Care Information Centre, also known as NHS Digital
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+import { Injectable } from '@angular/core';
+import {
+  CatalogueItemDomainType,
+  DataType,
+  Profile,
+} from '@maurodatamapper/mdm-resources';
+import { QueryBuilderConfig } from 'angular2-query-builder';
+import { catchError, forkJoin, map, Observable, of } from 'rxjs';
+import {
+  DataElementSearchResult,
+  DataRequestQueryPayload,
+  QueryExpression,
+} from '../data-explorer/data-explorer.types';
+import { ProfileService } from './profile.service';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class QueryBuilderService {
+  constructor(private profileService: ProfileService) {}
+
+  private getQueryBuilderDatatype(dataType?: DataType): Observable<Profile> {
+    if (dataType?.domainType === CatalogueItemDomainType.PrimitiveType) {
+      const requestOptions = {
+        handleGetErrors: false,
+      };
+
+      return this.profileService.get(
+        CatalogueItemDomainType.PrimitiveType,
+        dataType?.id ?? '',
+        'uk.ac.ox.softeng.maurodatamapper.plugins.explorer.querybuilder',
+        'QueryBuilderPrimitiveTypeProfileProviderService',
+        requestOptions
+      );
+    }
+    return of({} as Profile);
+  }
+
+  public setupConfig(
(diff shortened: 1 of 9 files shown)


src/app/mauro/query-builder.service.ts
```diff
@@ -0,0 +1,145 @@
...
+} from '../data-explorer/data-explorer.types';
+import { ProfileService } from './profile.service';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class QueryBuilderService {
+  constructor(private profileService: ProfileService) {}
+
+  private getQueryBuilderDatatype(dataType?: DataType): Observable<Profile> {
```
pjmonks, line 40: I'm surprised that `private` methods can go above `public` methods, I thought there was an eslint rule to enforce such ordering.

src/app/mauro/query-builder.service.ts
```diff
@@ -0,0 +1,145 @@
...
+      );
+    }
+    return of({} as Profile);
+  }
+
+  public setupConfig(
+    dataElements: DataElementSearchResult[],
+    query?: DataRequestQueryPayload
+  ): Observable<
+    [
```
pjmonks, line 61: Since this is quite a complex return type, it might be easier to read if it were a type alias e.g.

```ts
export type QueryBuilderConfigValues = [DataElementSearchResult[], Required<DataRequestQueryPayload> | undefined, QueryBuilderConfig]
```

Or would it not be more readable to just define an `interface`? The problem I find with tuples is once you get past 2 values they appear cumbersome.

src/app/mauro/query-builder.service.ts
```diff
@@ -0,0 +1,145 @@
...
+      );
+
+      const dataElement$ = of(dataElement);
+
+      return forkJoin([profile$, dataElement$]);
+    });
+
+    return forkJoin(requests$).pipe(
+      map((items) => {
+        const queryCondition = query
```
pjmonks, line 85: This `map()` operator is quite length, could the inner code be extracted to a separate private function to help readability?

src/app/mauro/query-builder.service.ts
```diff
@@ -0,0 +1,145 @@
...
+              rules: [],
+            };
+
+        const config: QueryBuilderConfig = {
+          fields: {},
+        };
+
+        items.forEach(([data, dataElement]) => {
+          const dataTypeString =
+            dataElement.dataType?.enumerationValues?.length ?? 0 > 0
```
pjmonks, line 98: The original code before your refactor had a separate function define the correct type string to use, could this be brought back? I'm thinking ahead to #169 where I need to also add another custom type string to handle model data types.

src/app/mauro/query-builder.service.ts
```diff
@@ -0,0 +1,145 @@
...
+              options: (dataElement.dataType?.enumerationValues ?? []).map(
+                (dataOption) => {
+                  return {
+                    name: dataOption.key,
+                    value: dataOption.value,
+                  };
+                }
+              ),
+              defaultValue:
+                dataTypeString.toLowerCase() === 'number'
```
pjmonks, line 117: Similar case with this default value selection, this could potentially grow bigger.

pjmonks (changes requested):
I haven't had a chance to test it properly yet, waiting for the `mdm-plugin-explorer` build to produce a snapshot with your related changes. I've listed some code feedback for now though which might be worth reviewing.

src/app/mauro/query-builder.service.ts
NigelPalmer: I have moved private methods to the bottom of the class.

src/app/mauro/query-builder.service.ts
NigelPalmer: I have added an interface. I am not a big fan of tuples myself.

src/app/mauro/query-builder.service.ts
NigelPalmer: I have refactored the class to break it down into sensibly named private methods to  make it easier to maintain.

src/app/mauro/query-builder.service.ts
NigelPalmer: This was done as part of the refactoring.

src/app/mauro/query-builder.service.ts
NigelPalmer: Private method added as suggested to maintain this.

src/app/pages/request-query/request-query.component.ts
```diff
@@ -91,14 +101,24 @@ export class RequestQueryComponent implements OnInit, IModelPage {
...
+        catchError(() => {
+          return this.errorResponse(
+            'There was a problem configuring your request queries.'
+          );
         })
       )
-      .subscribe(([dataElements, query]) => {
-        this.dataElements = this.mapDataElements(dataElements);
-        this.query = query;
+      .subscribe(([queryConfiguration]) => {
```
pjmonks, line 118: Linked to my comment about the return type - if just the `Observable<QueryConfiguration>` type was returned, then this destructuring wouldn't be necessary.

pjmonks (changes requested):
Code is now looking cleaner and finally got to test it, all seems functional and as expected.

My final feedback is to do with the return type on `QueryBuilderService.setupConfig()`, I think it needs adjusting.

src/app/mauro/query-builder.service.ts
```diff
@@ -0,0 +1,181 @@
...
+@Injectable({
+  providedIn: 'root',
+})
+export class QueryBuilderService {
+  constructor(private profileService: ProfileService) {}
+
+  public setupConfig(
+    dataElements: DataElementSearchResult[],
+    query?: DataRequestQueryPayload
+  ): Observable<[QueryConfiguration]> {
```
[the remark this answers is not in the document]
NigelPalmer: Updated as suggested.

src/app/pages/request-query/request-query.component.ts
NigelPalmer: Updated as suggested.

pjmonks (approved):
All good, great work 👍

PR #199, opened 2023-01-10
gh-188 Prevent removal of data elements from submitted requests

### LINKED ISSUE #188: Submitted data requests should not allow option to remove data elements
# Description

When viewing a submitted data request, it should not be possible to remove a data element from the request since it is now locked in place and represents the request at that point in time.

# Steps to reproduce

1. Add one or more data elements to any request
2. Go to the "My requests" screen and view that request.
3. Submit the request to change it to the "Submitted" state.

# Expected behavior

The "Remove" button listed in each data element row should not be visible. This would also apply to the multi-select actions too.

# Screenshots

image

pjmonks:
Resolves #188 

Submitted request:

image

Unsent request:

image

### DIFF
--- a/src/app/data-explorer/data-element-row/data-element-row.component.html
+++ b/src/app/data-explorer/data-element-row/data-element-row.component.html
@@ -36,6 +36,7 @@ SPDX-License-Identifier: Apache-2.0
     </span>
     <span class="mdm-data-element-row__header-button-panel">
       <button
+        *ngIf="canDelete"
         mat-raised-button
         class="mdm-data-element-row__header-button"
         color="primary"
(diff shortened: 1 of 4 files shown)


NigelPalmer (changes requested):
- [X] Remove button visible for unsent items.
- [X] Remove button not visible for submitted items.

Could we do anything to maintain the position of buttons after the "Removed" button is hidden? At the moment the "Add to other requests" button moves to the left after the "Remove" button is hidden. I think this weakens the user experience since users get used to the position of controls on the screen and muscle memory guides them to them.

pjmonks:
Updated the button alignment as requested.

Unsent:

image

Submitted:

image

NigelPalmer (approved):
- [X] The "Add to other request" button now stays in the same position when the "Remove" button is hidden.

**Passed**

PR #213, opened 2023-01-16
gh-201 footer to be displayed at the bottom

### LINKED ISSUE #201: Align footer to bottom of browser window
# Description

If the footer is visible it does not appear at the bottom of the browser window. This should always sit at the bottom of the browser window when visible and there should be no white space beneath it. This rule applies to all pages.

# Steps to reproduce

e.g.
1. Browse and bookmark a single data element.
2. Go to 'Bookmarks'
3. Footer is displayed half way up the page.

# Expected behavior

Footer should always be at the bottom of the browser window when it is visible.

# Screenshots
image

# Environment

## Desktop 

Please complete the following information:
- OS: Windows
- Browser: Seen in Edge and Chrome

CristianGimenez:
CSS changes:
Fixing the position to the bottom. 
But with dynamic content, only with that caused that when rendering extra content, the footer was out of place, to solve this, fixed the height of the footer, and added bottom padding in the body. Now the footer is in position even when the body is too small or some content is added after the window loaded.
Closes #201

### DIFF
--- a/src/app/app.component.html
+++ b/src/app/app.component.html
@@ -16,7 +16,7 @@ limitations under the License.
 
 SPDX-License-Identifier: Apache-2.0
 -->
-<div class="default-theme">
+<div class="default-theme page-container">
   <div class="mat-typography mdm-app-component">
     <mdm-header
       [logoLink]="logoLink"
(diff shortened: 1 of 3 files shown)


OButlerOcc (commented):
Footer is now always visible but I'm not sure that should be the case. I believe the footer should be at the bottom of every page but only visible when you scroll down to see it, as it is now it eats up a lot of screen. Would you like me to seek clarity on this as its a bit ambiguous from the task description?

pjmonks:
I haven't checked this PR yet, but the `mdm-ui` repo had a similar PR here: MauroDataMapper/mdm-ui#692. It would probably be similar, but the CSS/HTML might to change due to the different layouts.

jamesrwelch:
@OButlerOcc @CristianGimenez , yes, that is the intention, please, that it's at the bottom of every page, and that each page is at least the height of the screen, such that the footer never appears above the bottom of the page.

OButlerOcc:
All tested well except request detail where I noticed and fixed a padding issue (my fault)

OButlerOcc (approved):
Tested, works on all tested pages. 
Fixed a small issue with request detail (my fault)
Rebased on dev.

PR #215, opened 2023-01-17
gh-207: Request card clickable and ellipsis for 

### LINKED ISSUE #207: Make current data requests clickable
# Description

When browsing websites if I see a large box with rounded corners I expect to be able to click on it and something to happen. This is not the case on the MDM Explorer home page. At the bottom of the page is a section labelled "Your current data requests" with such boxes displayed. The user should be able to click on these boxes and be taken to the relevant "Requests" page. Then the "Manage my requests" button can be removed since it will become redundant. If a user wishes to see all their requests they can use the shopping cart icon in the top right. 

# Additional context

In the following screen shot the "COVID" and "Date Test" boxes should be clickable.
image

CristianGimenez:
Cards are now clickable and the bottom button is removed.

As per the ellipsis:

Cards have now fixed height. And I managed the amount of lines to be shown using [line-clamp webkit](https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-line-clamp)

Closes #207

### DIFF
--- a/src/app/pages/dashboard/dashboard.component.html
+++ b/src/app/pages/dashboard/dashboard.component.html
@@ -69,7 +69,10 @@ SPDX-License-Identifier: Apache-2.0
       <div *ngIf="currentUserRequests.length > 0">
         <p-carousel [value]="currentUserRequests" [numVisible]="3" [numScroll]="3">
           <ng-template let-request pTemplate="item">
-            <div class="mdm-dashboard__request-card">
+            <div
+              class="mdm-dashboard__request-card"
+              routerLink="/requests/{{ request.id }}"
+            >
               <h3>{{ request.label }}</h3>
               <p>{{ request.description }}</p>
             </div>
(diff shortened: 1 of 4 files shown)


src/app/pages/dashboard/dashboard.component.html
```diff
@@ -69,17 +69,15 @@ <h2>Your current data requests</h2>
       <div *ngIf="currentUserRequests.length > 0">
         <p-carousel [value]="currentUserRequests" [numVisible]="3" [numScroll]="3">
           <ng-template let-request pTemplate="item">
-            <div class="mdm-dashboard__request-card">
+            <div
+              class="mdm-dashboard__request-card"
+              routerLink="/requests/{{ request.id }}"
```
pjmonks, line 74: An equivalent is:

```html
<div [routerLink]="[ '/requests', request.id ]">
```

This is acceptable though.

pjmonks (approved):
I've tested this locally and it works as expected, and the code looks good. Well done 🏆 

Approved ✔️

PR #217, opened 2023-01-17
fixed status chip

OButlerOcc:
Status chip was not present in request detail page when description was blank,
noticed in demo.
Boolean state fix


pjmonks (approved):
Looks good to me 👍

PR #223, opened 2023-01-20
gh-204: Remove dataElements from queries when removing from request

### LINKED ISSUE #204: Remove field from query when removed from request
# Description

When writing a query the user can only select those fields that are included in the Request. If the user removes a field from the request that has been referenced by the query then an error will occur when trying to edit that query. 

# Steps to reproduce

1. Create a request with 2 data elements.
2. Create a query referencing one of those data elements.
3. Delete the data element that is referenced by the query.
4. Try and edit the query. Application error occurs.

# Expected behaviour

When a field is removed from the request it should be removed from the query.

CristianGimenez:
Added new method `deleteDataElementsFromQuery `on service `data-requests.service`.
Refactoring `removeItem `and `removeSelected `so both use the same code of `removeOneOrMoreDataElements `in `my-request-detail.component`

Adding and fixing tests for the new methods.

closes #204

### DIFF
--- a/src/app/data-explorer/data-requests.service.ts
+++ b/src/app/data-explorer/data-requests.service.ts
@@ -515,14 +515,14 @@ export class DataRequestsService {
    *
    * @param requestId The unique identifier of the data request.
    * @param type The type of query to get.
-   * @param dataElementLabel the label of the data element to remove from the query.
+   * @param dataElementLabels the labels of the data elements to remove from the query.
    * @returns a {@link DataRequestQueryPayload} with the query after removing the
    * data elements, or empty if the request or query cannot be found.
    */
   deleteDataElementsFromQuery(
     requestId: Uuid,
     type: DataRequestQueryType,
-    dataElementLabel: string
+    dataElementLabels: string[]
   ): Observable<DataRequestQueryPayload> {
     let updatedQuery: DataRequestQueryPayload;
 
@@ -534,9 +534,11 @@ export class DataRequestsService {
 
         updatedQuery = query;
 
-        updatedQuery.condition.rules = query.condition.rules.filter(
-          (item) => !(item as QueryExpression)?.field?.startsWith(dataElementLabel)
-        );
+        dataElementLabels.forEach((label) => {
+          updatedQuery.condition.rules = query.condition.rules.filter(
+            (item) => !(item as QueryExpression)?.field?.startsWith(label)
+          );
+        });
 
         return this.createOrUpdateQuery(requestId, updatedQuery);
       })
--- a/src/app/data-explorer/data-requests.service.ts
+++ b/src/app/data-explorer/data-requests.service.ts
@@ -43,7 +43,6 @@ import {
   switchMap,
   throwError,
   from,
-  empty,
 } from 'rxjs';
 import { UserDetails } from '../security/user-details.service';
 import { DataModelService } from '../mauro/data-model.service';
@@ -510,8 +509,7 @@ export class DataRequestsService {
   }
 
   /**
-   * Removes all rules containing the data element label. And returns
-   * the new value of the query.
+   * Removes all data labels found in a query attached to a data request.
    *
    * @param requestId The unique identifier of the data request.
    * @param type The type of query to get.
--- a/src/app/data-explorer/data-requests.service.ts
+++ b/src/app/data-explorer/data-requests.service.ts
@@ -43,6 +43,7 @@ import {
   switchMap,
   throwError,
   from,
+  empty,
 } from 'rxjs';
 import { UserDetails } from '../security/user-details.service';
 import { DataModelService } from '../mauro/data-model.service';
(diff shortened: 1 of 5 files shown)


src/app/pages/my-request-detail/my-request-detail.component.ts
```diff
@@ -277,30 +244,58 @@ export class MyRequestDetailComponent implements OnInit {
...
-          result.failures.forEach((item: DataElementOperationResult) =>
-            console.log(item.message)
-          );
+      .subscribe((resultMultiple: DataElementMultipleOperationResult) => {
+        const labels = resultMultiple.successes.map((success) => {
+          return success.item.label;
+        });
+
+        // Remove data element from request queries.
+        this.removeDataElementFromQuery(labels, this.dataQueryType);
```
pjmonks, line 259: Each of these `removeDataElementFromQuery()` calls actually creates a nested `subscribe()` call too. This is considered an anti-pattern in Angular/RxJs because now the operations become non-deterministic, at worst producing race conditions. You can kind of see the effect a bit in the UI when it refreshes - the loading screen disappears once the element is removed, but maybe a second or two later the query components re-render when their operations complete.

It is normally better to have all these observables run together in the same `pipe()` and have one `subscribe()` at the end which handles the final success case. This makes everything deterministic, even if there are potential delays e.g. with network calls to complete. At least you can then guarantee that each steps runs in the exact order you expect.

A cut down version of this then might be:

```ts
removeOneOrMoreDataElements(items: DataElementSearchResult[]) {
  this.okCancelItemList(items)
    .afterClosed()
    .pipe(
      switchMap((result: OkCancelDialogResponse) => {
        // Show UI we're doing something...
        this.broadcast.loading({ loading: true });
        // Perform action to remove data element...
        return this.dataRequests.deleteDataElementMultiple(items, dataModel);        
      }),
      switchMap((resultMultiple: DataElementMultipleOperationResult) => {
        // Perform two async actions but wait for both to finish...
        return forkJoin([
          this.dataRequests.deleteDataElementsFromQuery(this.request.id, 'data', items),
          this.dataRequests.deleteDataElementsFromQuery(this.request.id, 'cohort', items),
        ])
      }),
      finalize(() => {
        // Always stop the "loading" overlay, even if an error occurs...
        this.broadcast.loading({ loading: true });
      })
    )
    .subscribe(() => {
      // Display success messages here...
    })
}
```

This probably won't compile straight up, and you may need to pass the results of each action down the pipe to be used later, but this would be the general idea.

src/app/pages/my-request-detail/my-request-detail.component.ts
```diff
@@ -257,18 +224,18 @@ export class MyRequestDetailComponent implements OnInit {
     }
   }
 
-  removeSelected() {
-    const itemList = this.requestElements.filter((item) => item.isSelected);
-    this.okCancelItemList(itemList)
+  removeOneOrMoreDataElements(items: DataElementSearchResult[]) {
```
pjmonks, line 227: The refactoring of the remove element code into a common function is generally good. It could probably even be made into a `private` function 👍

pjmonks (changes requested):
When testing it via the UI it definitely works. I'm genuinely impressed with the level you're at with understanding RxJs, observables and the tests given your short amount of time with them 🏆 

The one thing I'd like you to revisit is the nested `subscribe()`` issue I explained in my feedback, I think this could be improved before I approve this.

pjmonks:
One other thing I noticed is that there are eslint errors found, which prevents the Jenkins build from completing.

A handy npm command we have is `npm run pr-checks` which carries out the same checks as the CI build but locally - check license headers, eslint and tests. Running this before submitting a PR is usually good practice to make sure obvious issues are found first.

From the issues I've seen, your eslint errors shouldn't be too hard to fix.

src/app/data-explorer/data-requests.service.ts
```diff
@@ -506,4 +508,40 @@ export class DataRequestsService {
       })
     );
   }
+
+  /**
+   * Removes all rules containing the data element label. And returns
```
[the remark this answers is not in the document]
CristianGimenez: Changed as suggested

src/app/pages/my-request-detail/my-request-detail.component.ts
CristianGimenez: Refactored to solve antipattern, thanks

src/app/pages/my-request-detail/my-request-detail.component.ts
CristianGimenez: made it private

CristianGimenez:
Thanks for your kind words and your thoughtful comments, I've pushed a new commit which should address all your comments. Could you review again @pjmonks ?

pjmonks (approved):
Much better 👍 Approved ✔️

PR #225, opened 2023-01-23
gh-219 Display "value" of enumeration in query builder category fields

### LINKED ISSUE #219: Enum dropdown in QueryBuilder should display "name" rather than "value"
# Description

When editing a "category" data element in the query builder, the dropdown shows its numeric value whereas the query data section shows the name of the category (see image). It would be more user friendly to display the name in the dropdown.

# Steps to reproduce

e.g.
1. Create a new request which include some dataelement of type category
2. Create either a data query or a cohort query (it happens in both places)
3. Add a new rule and select the category element as field.
4. After selecting the operator, the next dropdown will show numbers instead of user friendly names.

# Expected behavior

The dropdown show friendly names and not values

# Screenshots

image

pjmonks:
image

Fixes #219 

It should be noted that the change actually uses an enumeration's value, not the "name" (or "key"), since the keys used for some of the OUH model use numbers while the "value" uses a display string - most times both are the same. It should be confirmed if this is correct.

### DIFF
--- a/src/app/data-explorer/querybuilder/querybuilder.component.html
+++ b/src/app/data-explorer/querybuilder/querybuilder.component.html
@@ -157,7 +157,7 @@ SPDX-License-Identifier: Apache-2.0
         >
           <mat-select [(ngModel)]="rule.value" (ngModelChange)="onChange()">
             <mat-option *ngFor="let opt of options" [value]="opt.value">
-              {{ opt.name }}
+              {{ opt.value }}
             </mat-option>
           </mat-select>
         </mat-form-field>


CristianGimenez:
Some of the values aren't really very user friendly, but that depends on the data. At least now the queries display the same thing that the user can see in the dropdown, I think that is much better.
image

CristianGimenez (approved):
Sorry, I wrote a comment but apparently that does not count as review. Now the dropdown and the query builder show the same value, which is great.

PR #226, opened 2023-01-24
gh-206 Fix date formatting bug in meql-pipe

### LINKED ISSUE #206: Date format changing when saving a query
# Description

When defining a date field in the query builder the date will me displayed as 'dd/MM/yyyy' in the MEQL. After saving the query this will change to it's raw form. 'yyyy-MM-ddTHH:mm:ss.SSSZ'

# Steps to reproduce

1. Create a request that references a data element that is a date type.
2. Create a Cohort query and reference the date field. 
3. Save the changes.
4. The raw format will be displayed in the MEQL.

# Expected behavior

The date shown in the MEQL window should always be in the format 'dd/MM/yyyy'

# Screenshots

Before save:
image

After save:
image

pjmonks:
Date values could be moment instances _or_ JSON date strings. Must check both cases to ensure a value can be formatted as a date.

Fixes #206

### DIFF
--- a/src/app/app.module.ts
+++ b/src/app/app.module.ts
@@ -41,7 +41,12 @@ import { DataExplorerService } from './data-explorer/data-explorer.service';
 import { Observable } from 'rxjs';
 import { USER_IDLE_CONFIGURATION } from './external/user-idle.service';
 import { LOCALE_ID } from '@angular/core';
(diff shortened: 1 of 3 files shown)


OButlerOcc (approved):
Changes make sense, tested and correctly works

PR #227, opened 2023-01-24
gh-222: Dynamic size cards in carousel

### LINKED ISSUE #222: Request Cards are too big when description is missing or not long enough to fill card
# Description

The clickable request cards on the home page are too big when a description is not provided. This is creating a lot of empty white space that looks odd.

# Steps to reproduce

1. Create some requests and do not add any descriptions.
2. View the home page. The request cards look unnecessarily tall.

# Expected behavior

1. The maximum height of the card should be the height it is currently set at in the screenshot below.
3. If there is not enough text to fill the card then reduce the card height so the text fills it.
4. The height of all cards in the carousel should be set to equal the tallest card in the carousal so that they are all the same height. This potentially could create the white space issue seen in the screenshot below, since the card that is full of text may not be visible, but this is acceptable.

# Screenshots
image

CristianGimenez:
As mentioned in #222 right now the height of each carousel card are hardcoded and that can produce lots of empty space that seems odd.

This change has some calculations added when we load the requests for the carousel to check the length of the longest description in the requests and calculate how many lines approx the height now is dynamic.

Note: depending on the font and the combination of letters not always 10 characters occupy the same space (expect in monospace fonts, so a lot of these calculations are only based on tests and their are still only approximate values)

example: 16 characters in each line:
VAVAVAVAVAVAVAVA
MMMMMMMMMMMMMMMM

Closes #222

### DIFF
--- a/src/app/pages/dashboard/dashboard.component.html
+++ b/src/app/pages/dashboard/dashboard.component.html
@@ -72,9 +72,12 @@ SPDX-License-Identifier: Apache-2.0
             <div
               class="mdm-dashboard__request-card"
               routerLink="/requests/{{ request.id }}"
+              [ngStyle]="{ height: currentCardsHeightString }"
             >
               <h3>{{ request.label }}</h3>
-              <p>{{ request.description }}</p>
+              <p [ngStyle]="{ '-webkit-line-clamp': itemCardNumerOfLinesToShow }">
+                {{ request.description }}
+              </p>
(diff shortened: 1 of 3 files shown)


pjmonks (approved):
Very clear code and well documented, works as defined. If this were necessary in more than one place I'd suggest moving this line calculation code to a function/service, but no need right now.

Approved 👍

Issue #209, opened 2023-01-11
Make the request query editor page consistent with the rest of the request management pages


NigelPalmer:
# Description

The request query editor page should be styled to look more like the other request management pages. I.e it should have a dark green bar and sandy coloured background. At the moment it is plain white. The positioning of the name of the request and the badge also needs to be considered. See #208  that has been raised about wider reaching badge alignment issues.

# Additional context

Cohort Query Editor, lacking branding:
![image](https://user-images.githubusercontent.com/110250803/211859071-222db4d7-fdb0-47ad-b231-f3a4708ca027.png)

Cohort Query presented on details page. The editor needs to be branded to adhere to this scheme:
![image](https://user-images.githubusercontent.com/110250803/211858920-4056003f-3f10-4845-a607-7c83c5439279.png)

pjmonks:
I am making some changes in #203 for consistency which may help with this.

PR #229, opened 2023-01-24
Gh-209 Updated css for query builder and cleaned up a bit

OButlerOcc:
Updated the query builder styling to match its host files.
Cleaned up a few ugly css stylings in host file while I was poking through

### DIFF
--- a/src/app/pages/request-query/request-query.component.html
+++ b/src/app/pages/request-query/request-query.component.html
@@ -42,9 +42,11 @@ SPDX-License-Identifier: Apache-2.0
             <button
               *ngIf="status === 'ready'"
               mat-raised-button
+              color="primary"
               [disabled]="!dirty"
               (click)="save()"
             >
+              <span class="fa-solid fa-paper-plane"></span>
               Save changes
             </button>
             <div class="mdm-request-query__header-name"></div>
--- a/src/app/pages/request-query/request-query.component.html
+++ b/src/app/pages/request-query/request-query.component.html
@@ -16,43 +16,55 @@ limitations under the License.
 
 SPDX-License-Identifier: Apache-2.0
 -->
-<div *ngIf="dataRequest" class="container">
-  <div class="main-row hero">
-    <h1>Request Query</h1>
-  </div>
-  <div class="mdm-request-query__back-link">
-    <a [routerLink]="[backRouterLink, backRouterRequestId]">
-      <span class="fa-solid fa-arrow-left"></span>
-      {{ backLabel }}
-    </a>
-    <button
-      *ngIf="status === 'ready'"
-      mat-flat-button
-      color="primary"
-      [disabled]="!dirty"
-      (click)="save()"
-    >
-      Save changes
-    </button>
-  </div>
-  <div *ngIf="status === 'loading'" class="main-row mdm-request-query__loading">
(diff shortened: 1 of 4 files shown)


OButlerOcc:
Edited a little to make it match petes work, cleaned up my mistake in making a misnamed remote/local branch. Ready for rereview

src/app/pages/request-query/request-query.component.html
```diff
@@ -16,43 +16,57 @@
...
+
+    <div class="col">
+      <div class="mdm-request-query__header">
+        <div class="mdm-request-query__header-title">
+          <div class="mdm-request-query__header-buttons">
+            <button
+              *ngIf="status === 'ready'"
+              mat-raised-button
+              color="primary"
+              [disabled]="!dirty"
```
pjmonks, line 46: When the "Save changes" button is disabled, it doesn't look great given the background colour:

image

I think it would look better if the button was just hidden completely with `*ngif` than disabled.

src/app/pages/request-query/request-query.component.html
```diff
@@ -16,43 +16,57 @@
...
+<div class="container">
+  <div *ngIf="dataRequest">
+    <div class="main-row hero">
+      <h1>{{ dataRequest.label }}</h1>
+    </div>
+    <div *ngIf="status === 'loading'" class="mdm-my-request__progress">
+      <mat-progress-bar color="primary" mode="indeterminate"></mat-progress-bar>
+    </div>
+    <div class="mdm-request-query__back-link">
+      <a [routerLink]="[backRouterLink, backRouterRequestId]">
```
pjmonks, line 28: Maybe replace this with the new `mdm-back-link` component for consistency?

src/app/pages/request-query/request-query.component.html
```diff
@@ -16,43 +16,57 @@
...
+        <mdm-request-status-chip [status]="dataRequest.status"></mdm-request-status-chip>
+      </div>
+      <div *ngIf="!dataRequest.description" class="mdm-request-query__description">
+        <p>No description provided</p>
+        <mdm-request-status-chip [status]="dataRequest.status"></mdm-request-status-chip>
       </div>
     </div>
-    <div *ngIf="dataElements">
+
+    <div *ngIf="dataElements" class="mdm-request-query__query-editor">
```
pjmonks, line 69: I think it would nicer if there were some spacing between the main request description and the query builder, would look less squashed/compressed:

image

pjmonks (changes requested):
Mostly looking good. Could you check over the feedback I've given please before I approve it?

src/app/pages/request-query/request-query.component.html
OButlerOcc: Hiding it squished the green bar so it's opactity 0 when disabled now

src/app/pages/request-query/request-query.component.html
OButlerOcc: Used the same padding value as on the request detail page.

src/app/pages/request-query/request-query.component.html
OButlerOcc: Fixed

pjmonks (approved):
Approved, looks nicer 👍

PR #233, opened 2023-01-30
gh-158 Add message and links for adding more data elements to requests

### LINKED ISSUE #158: Request List: Cannot add Data Point
On the list of Requests it is not possible to add a new data point to an unsent request. The user has to go to Search or Browse using the menu bar at the top of the page and then choose to add new data points. I think it would be good to have two buttons next to the "Remove selected..." button that say "Add using browse" and "Add using search" which when clicked would take the user to the respective screens. Even though this would just do the same as the menu items at the top of the page, I think this would make the user experience smoother and more intuitive. Or at the very least, a label saying "Add new items using the Browse and Search options at the top of the page"

pjmonks:
image

Resolves #158


NigelPalmer (approved):
Perfect! Nice and simple solution that provides clear instructions and helpful links to the user.

PR #234, opened 2023-01-31
Feature/gh 230 time selector misaligned

OButlerOcc:
Fix for time control being broken, 
highlighted an issue with some data type controls. making a separate task to fix them.


pjmonks (approved):
image

Alignment looks fixed, nice one 👍

PR #248, opened 2023-02-03
Date time fix and fix to operators

OButlerOcc:
Required: https://github.com/MauroDataMapper-Plugins/mdm-plugin-explorer/pull/17

Added datetime data type
Fixed missing operators for time, datetime and others

to test locally.

Clone, build and publish to local maven the mdm plugin explorer
Start mdm-application-build with a reference to plugin-explorer (should already be there)
Open mdm-ui add a datetime type and other missing types
Test in research browser

Nigel should probably be the one to check the operators block works the way I think it does (seems to work in testing)

### DIFF
--- a/src/app/data-explorer/querybuilder/querybuilder.component.ts
+++ b/src/app/data-explorer/querybuilder/querybuilder.component.ts
@@ -50,8 +50,8 @@ export class QueryBuilderComponent implements OnInit {
   @Output() queryChange = new EventEmitter<QueryCondition>();
 
   operatorMap: { [key: string]: string[] } = {
-    //non-default data types require specifying operators in order to populate the operator drop down.
-    //Is also used to extend some default operators
+    // non-default data types require specifying operators in order to populate the operator drop down.
+    // Is also used to extend some default operators
     /* eslint-disable */
     string: ['=', '!=', 'contains', 'like', 'startswith', 'endswith'],
     datetime: ['=', '!=', '>', '<', '=>', '=<'],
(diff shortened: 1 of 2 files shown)


src/app/data-explorer/querybuilder/querybuilder.component.ts
```diff
@@ -52,6 +52,11 @@ export class QueryBuilderComponent implements OnInit {
   operatorMap: { [key: string]: string[] } = {
     /* eslint-disable-next-line */
     string: ['=', '!=', 'contains', 'like', 'startswith', 'endswith'],
+    datetime: ['=', '!=', '>', '<', '=>', '=<'],
+    date: ['=', '!=', '>', '<', '=>', '=<'],
+    time: ['=', '>', '<', '=>', '=<'],
+    boolean: ['=', '!='],
+    number: ['=', '!=', '>', '<', '=>', '=<'],
```
NigelPalmer, line 59: The Query Builder has default operators for it's default types. Therefore, it is only necessary to add operators here when the defaults are to be overridden or the type is not a default Query Builder type. This means that line 55 (datetime) should be retained and lines 56-59 (date, time, boolean and number) can be deleted since they are not required.

Please also add a comment to the code explaining this to avoid confusion in the future.

NigelPalmer (changes requested):
Functionally this is working correctly. The Time and DateTime fields are being displayed with appropriate controls. The Time != operator was missing, but reverting to defaults will resolve this. (see code comments)

Reopening so minor code tidy-up can be carried out.

NigelPalmer (approved):
All looks good now.

pjmonks (approved):
@NigelPalmer has already reviewed this, so I will approve also after latest commits were pushed to build the branch.

PR #252, opened 2023-02-07
gh-240 Add sort control to Request List pages

### LINKED ISSUE #240: Add ordering options to Request lists
# Description

gh-202 added ordering to sort lists of Requests by name in ascending order to make them easier to manage. This sorting could be improved to allow other sorting options. These options would be:

1. Sort by name ascending (already implemented)
2. Sort by name descending.
3. Sort by created date ascending.
4. Sort by created date descending

Ideally the sort option would be remembered so that the next time the user visits that screen they would see the list in the same situation as they last viewed it. This should be remembered even if the user logs out and in again.

# Additional context

A control will be required to allow the user to choose their sorting option.

pjmonks:
- Updated "My Requests" and "Templates" pages
- Can sort by label in ascending or descending order
- Updated the "My Request" filter too

Resolves #240 

# Limitations

All sorting must be done on the client-side because the Mauro endpoints used for fetching the lists do not support sort arguments. Also no date fields are returned in the list objects, meaning `label` is currently the only way to sort requests.

# Screenshots

image

image

### DIFF
--- a/src/app/mauro/sort.type.ts
+++ b/src/app/mauro/sort.type.ts
@@ -1,3 +1,5 @@
+import { SortOrder } from '../data-explorer/data-explorer.types';
+
 /*
 Copyright 2022-2023 University of Oxford
 and Health and Social Care Information Centre, also known as NHS Digital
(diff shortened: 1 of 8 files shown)


NigelPalmer (changes requested):
Functionally all works and code changes look good. My only complaint is that the filter and the sort are on opposite ends of the Request list. I think it would be a better user experience if these were next to each other as so:
image

pjmonks:
@NigelPalmer I've positioned the filter and sort controls to the right now, as you've suggested.

NigelPalmer (approved):
Happy with latest changes. All good.

PR #265, opened 2023-02-14
gh-262 Add "Open request" option to request created/updated dialogs

### LINKED ISSUE #262: Add "Open request" option when elements are successfully added to requests
# Description

Once one or more elements have been added to a data request successfully, the user should be able to open that request directly from the success dialog.

# Additional context

image

This is the current dialog. Adding an extra button called "Open request" which links directly to that request would be useful. The other buttons should remain as is.

pjmonks:
Allows the user to jump straight to the request if they choose.

Resolves #262 

image


NigelPalmer (changes requested):
- [X] Search Results Page: "Open request" option displayed after assigning a field to an existing request
- [X] Search Results Page: "Open request" option displayed after assigning a field to a new request.
- [X] Request Page: "Open request" option displayed after assigning a field to an existing request
- [ ] Request Page: "Open request" option NOT displayed after assigning a field to a new request. This is different behaviour to the "Search results" page. I don't think it is caused by the changes in this task but I would expect the behaviour to be the same. **Please explain or fix or create follow-up task.**
image

NigelPalmer (approved):
Missing open request button now present. All working as expected. Passed.

PR #267, opened 2023-02-14
gh-246 Upgrade to node 16.10.0

### LINKED ISSUE #246: Upgrade to Node 16 with updated npm
# Description

Consider upgrading NodeJS to the next LTS version, which should also include an upgrade to npm.

# Additional context

Currently, the repo targets Node 14.18.1 (as specified by the `.nvmrc` file). This is still acceptable for Angular 14, but may become an issue for further upgrades in the future. According to https://unpkg.com/browse/@angular/core@14.2.12/package.json, Angular 14 supports:

```
"engines": {
    "node": "^14.15.0 || >=16.10.0"
  },
```

Suggestion is to upgrade to NodeJS 16.x.x at least.

Another reason to upgrade is that warnings can appear when running `npm install` regarding `package-lock.json` files being different versions. Ensuring the npm is in line with a later node version, this warning should then disappear.

# Actions

- [ ] Update `.nvmrc` file
- [ ] Install Node 16.x.x and update npm
- [ ] Test to ensure everything still builds and works

pjmonks:
Resolves #246


NigelPalmer (approved):
Everything is working as it should and the changes look correct. Passed.

PR #275, opened 2023-02-16
gh-271 Add entity selection support to query builder

### LINKED ISSUE #271: Query Builder: Group fields by schema and class, and provide additional information.
# Description

The Query Builder needs to be enhanced so that:
1. When choosing a field the fields are grouped by schema and class. This could be down with separate combo boxes for the schema and class, but this will make the UI quite wide. Have a look at https://zebzhao.github.io/Angular-QueryBuilder/demo/ and turn on the 'Entity Mode'. The other alternative is that the option has some sort of text above it displaying the schema and class, but again, this could make the UI quite wide. This will need to be played around with to produce something sensible.
2. After selecting the element, it should display the full path to the field in the query: e.g: "Diagnosis.Diagnosis (ARIA MedOnc).Diagnosis date time" = "15/02/2023".
3. When a field name is hovered over, a toast pop-up appears with the description of the field in it. This should be truncated to a sensible length with ellipsis at the end. On initial look it would appear that setupQueryFieldFromMapping in the query-builder.service.ts would be a good place to setup the description.

pjmonks:
- Map schema/class levels to each field to identify entities used
- Output full field name through MEQL pipe transform

Resolves #271 

image


NigelPalmer (approved):
- [X] Schema/class dropdown appears and filters elements to be selected.
- [X] Full field name displayed in query.

PR #280, opened 2023-02-20
gh-279: Query Builder: Display description information when hovering …

### LINKED ISSUE #279: Query Builder: Display description information when hovering over a field name
# Description

In the Query Builder, when expanding the field selection dropdown and hovering over a field name, a toast pop-up needs to appear with the description of the field in it. This should be truncated to a sensible length with ellipsis at the end.  If a description is not available then the text, "No additional information available" should be displayed.

# Additional context

This was originally included in #271 , but at the time looked like it wasn't possible. Building on the work done in #271, this can be done by using the entity paths to identify the correct DataElementSearchResult record, which holds the required description.

NigelPalmer:
…over a field name

Resolves #279 .

### DIFF
--- a/src/app/data-explorer/querybuilder/querybuilder.component.html
+++ b/src/app/data-explorer/querybuilder/querybuilder.component.html
@@ -123,6 +123,9 @@ SPDX-License-Identifier: Apache-2.0
             <mat-option
               *ngFor="let field of getFields(rule.entity)"
               [value]="field.value"
+              [matTooltip]="elementDescription(rule.entity, field.value)"
+              matTooltipClass="mdm-selection-item-tooltip"
+              matTooltipPosition="right"
             >
               {{ field.name }}
             </mat-option>
--- a/src/app/data-explorer/querybuilder/querybuilder.component.html
+++ b/src/app/data-explorer/querybuilder/querybuilder.component.html
@@ -123,7 +123,7 @@ SPDX-License-Identifier: Apache-2.0
             <mat-option
               *ngFor="let field of getFields(rule.entity)"
               [value]="field.value"
-              [matTooltip]="elementDescription(rule.entity, field.value)"
+              [matTooltip]="descriptions[rule.entity + '.' + field.value]"
               matTooltipClass="mdm-selection-item-tooltip"
               matTooltipPosition="right"
             >
--- a/src/app/data-explorer/querybuilder/querybuilder.component.spec.ts
+++ b/src/app/data-explorer/querybuilder/querybuilder.component.spec.ts
@@ -32,13 +32,17 @@ import { MeqlOutputComponent } from '../meql-output/meql-output.component';
 import { MatCard } from '@angular/material/card';
 import { QueryCondition } from '../data-explorer.types';
 import { mapModelDataTypeToOptionsArray } from '../query-builder.service';
-import { CatalogueItemDomainType } from '@maurodatamapper/mdm-resources';
+import {
+  CatalogueItemDomainType,
+  MdmResourcesConfiguration,
+} from '@maurodatamapper/mdm-resources';
 import { of } from 'rxjs';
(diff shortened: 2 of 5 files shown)


src/app/data-explorer/querybuilder/querybuilder.component.html
```diff
@@ -123,6 +123,9 @@
             <mat-option
               *ngFor="let field of getFields(rule.entity)"
               [value]="field.value"
+              [matTooltip]="elementDescription(rule.entity, field.value)"
```
pjmonks, line 126: I see a potential performance issue here because the template is calling a function to do input binding - but this is happening anyway because the `angular-querybuilder` library component is doing the same. The issue is that Angular change detection can no longer tell if this input binding has changed or not, so it is always going to be called and re-rendered. Confirm this by putting a `console.log()` line inside `elementDescription()` and viewing the results in the browser console.

The only way I can think of getting around this is to pre-cache all the descriptions. Here is some code I've tested which might work:

**querybuilder.component.ts**
```ts
descriptions: { [field: string]: string } = {};

ngOnInit() {
  this.dataElements.forEach((element) => {
      const entity = this.queryBuilderService.getEntity(element);
      const fullName = `${entity}.${element.label}`;

      const description = element.description
        ? element.description.length > 300
          ? element.description.substring(0, 300)
          : element.description
        : 'No additional information available';

      this.descriptions[fullName] = description;
    });
}
```

**querybuilder.component.html**
```html
<!-- Field Selector -->
      <ng-container
        *queryField="
          let rule;
          let fields = fields;
          let onChange = onChange;
          let getFields = getFields
        "
      >
        <mat-form-field
          appearance="outline"
          class="query-builder-field mat-form-field-no-padding"
        >
          <mat-select [(ngModel)]="rule.field" (ngModelChange)="onChange($event, rule)">
            <mat-option
              *ngFor="let field of getFields(rule.entity)"
              [value]="field.value"
              [matTooltip]="descriptions[rule.entity + '.' + field.value]"
              matTooltipClass="mdm-selection-item-tooltip"
              matTooltipPosition="right"
            >
              {{ field.name }}
            </mat-option>
          </mat-select>
        </mat-form-field>
      </ng-container>
```

Notice something similar is happening with `termSearchResults`. I'm not entirely sure if fetching from an array has the same affect as calling a function or not, but in general I think the change detection issue is avoidable if all data for binding is known up front.

src/app/data-explorer/querybuilder/querybuilder.component.spec.ts
```diff
@@ -47,6 +51,10 @@ describe('QueryBuilderComponent', () => {
           provide: TerminologyService,
           useValue: terminologyStub,
         },
+        {
+          provide: MdmResourcesConfiguration,
+          useValue: mdmResourcesConfiguration,
```
pjmonks, line 56: Providing `MdmResourcesConfiguration`  as a value to the `TestBedModule` looks like it is trying to get the tests running in any way possible. I think the tests should be updated to reflect the new functionality of getting the descriptions instead.

pjmonks (changes requested):
In essence the feature looks good. My main concern is with the way the change detector will always be running though - could you look into this please? If it is too much of an issue we could just accept it as it is.

src/app/data-explorer/querybuilder/querybuilder.component.spec.ts
NigelPalmer: I have added a test to make sure that descriptions are mapped correctly.

src/app/data-explorer/querybuilder/querybuilder.component.html
NigelPalmer: Updated as suggested. Yes, there was an issue with the old code constantly firing.

pjmonks (approved):
This seems to function the same and (I think) will perform better. I like how the descriptions appear actually, nice one 👍

PR #283, opened 2023-02-20
gh-277 Do not allow query editing if no data elements are available

### LINKED ISSUE #277: Query page hangs when there are no elements in a request
# Description

Query page hangs after a user clicks the "Create" query button when there are no elements in a request.

# Steps to reproduce

1. Create a new request with a single element
2. Delete the element from the request
3. Click on "Create" for either the Cohort Query or Data Query.
4. The Query page is displayed and the progress bar will be continuously moving. 

# Expected behavior

Hide the "Create" query button when they are no elements assigned to the request.

# Screenshots

image

pjmonks:
- Check if request is "empty" and if so hide the create/edit query buttons
- Modify request query page to auto redirect back to the request if loading and no data elements were found
- Test that request-query page redirects when no data elements available

Fixes #277

### DIFF
--- a/src/app/pages/my-request-detail/my-request-detail.component.html
+++ b/src/app/pages/my-request-detail/my-request-detail.component.html
@@ -44,14 +44,14 @@ SPDX-License-Identifier: Apache-2.0
     <mdm-data-query-row
       queryType="cohort"
       [condition]="cohortQuery"
-      [readOnly]="request.status === 'submitted'"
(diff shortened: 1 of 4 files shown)


OButlerOcc (approved):
Clean Changes, Works as expected with one caveat.
Unable to test if it works without refreshing the request page. 
Please see https://github.com/MauroDataMapper/mdm-explorer/issues/284
But it looks like an unrelated bug.

PR #289, opened 2023-02-22
gh 257: Fix callstack bug in big requests

### LINKED ISSUE #257: Request Detail: Maximum call stack size exceeded
# Description

If a request is created from the ICU schema then a "Maximum call stack size exceeded" error occurs when viewing the request detail.

# Steps to reproduce

1. Go to 'Browse'
2. Click on 'ICU' in the 'Schemas' list
3. Hover your mouse over ICU and wait for a button with text '...' to appear.
4. Click the '...' button.
5. Select 'Copy to new request'
6. Type in a request name an click ok.
7. Wait for the request to be created.
8. Click on 'View requests'
9. Click on the name of the request you just created.
10. The 'Maximum call stack size exceeded' error will appear.

# Expected behavior

All requests should be able to be opened without an error occurring.

# Screenshots

image

# Environment

## Desktop 

Please complete the following information:
- OS: Windows 10
- Browser: Edge
- Version: 109.0.1518.78

# Additional context

We are going to remove the feature to create a Request from an entire schema, but we still need to get to the bottom of this problem. A whole schemas worth of data elements could be added to the request on a class by class basis to request the whole schema.

CristianGimenez:
The main problem was that change events on the data elements row was emitting events that caused the father components to change the item within the data element component causing the change event to emit again causing an infinite loop.

I have refactored how it works, data element component now uses ngModelChange which only triggers when the users change it, so no update from its father will trigger it.

Data class and data schema components check now is based on if all its children are selected or not. This gets updated by handling the event from the children each time there is a change of value.

Updated the tests to accommodate the new workflow, but it should not cause the components to dramatically change its behaviour, just the way of achieving such behaviour.

closes #257

### DIFF
--- a/src/app/data-explorer/data-class-row/data-class-row.component.html
+++ b/src/app/data-explorer/data-class-row/data-class-row.component.html
@@ -23,10 +23,10 @@ SPDX-License-Identifier: Apache-2.0
         <mat-checkbox
           *ngIf="canDelete"
           class="mdm-data-class-row__header-title-checkbox"
-          [checked]="dataClassWithElements?.dataClass?.isSelected"
-          (change)="classChecked()"
+          [(ngModel)]="(dataClassWithElements?.dataClass)!.isSelected"
+          (ngModelChange)="ngModelOnChange()"
         ></mat-checkbox>
-        <h3>{{ dataClassWithElements?.dataClass?.label }}</h3>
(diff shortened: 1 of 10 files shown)


pjmonks (approved):
I can't find any fault with this 🎉 Tested this myself with a large request, it loads perfectly now. The code changes are well laid out and commented to explain what is happening, and tests are included.

Approved ✔️

PR #290, opened 2023-02-22
gh-285 Fix loading screen bug when removing items from a request

### LINKED ISSUE #285: Requests: Delete item is hanging page
# Description

When a schema, class or data element is deleted from a request a progress animation is displayed to the user. This animation remains on the screen and never goes away. If the user presses F5 then the screen refreshes and the items that were deleted disappear.

# Steps to reproduce

1. Create a new request and add some elements to it.
2. Go to the Request screen
3. Select the Request that you just created.
4. Click the bin icon on a data element. Web page appears to hang.
5. Press F5 to refresh page. Data element that was selected to be deleted will have disappeared.
6. Click the bin icon on a class. Web page appears to hang.
7. Press F5 to refresh page. Data class that was selected to be deleted will have disappeared.
8. Click the bin icon on a schema. Web page appears to hang.
9. Press F5 to refresh page. Data schema that was selected to be deleted will have disappeared.

# Expected behavior

When an item is deleted the progress animation should only be displayed whilst the UI is communicating with the backend and should disappear once the task is completed.

# Screenshots
image

pjmonks:
When checking existing queries to remove the same elements that were just removed, if the queries did not exist yet then the observable pipe returned EMPTY which short-circuited the observable stream to not pass the final value.

Fixed this by returning undefined instead and doing null checks where appropriate.

Fixes #285


greg-at-occ (commented):
I wonder if we could also put `this.broadcastService.loading()` from line 284 into the complete handler of `subscribe` so that it fires whether or not the inner observable emits, and also if an error is thrown.

It's also a bit confusing to have the loading animation started and stopped in different functions. Could we pull the processing of the `OkCancelDialogResult` out of `this.deleteItem` and bring them both into the same scope?

pjmonks:
@greg-at-occ I'm not sure I understood your first comment but think it was all related to the changes I've pushed which shuld hopefully simplify things a bit.

- Keep loading broadcast messages (on/off) within same scope of function
- Dialog ok/cancel response is handled in the observable pipe now where it is clear it is the first thing to check
- deleteItem() switch statement has been simplified with less indentation

greg-at-occ (approved):

PR #295, opened 2023-02-27
gh-274 Load theme settings from Mauro

### LINKED ISSUE #274: Add Theming to Mdm Explorer: Load colours and fonts from backend
# Depends on

* #264
* MauroDataMapper-Plugins/mdm-plugin-explorer#18
* MauroDataMapper-Plugins/mdm-plugin-explorer#19

# Description

Load api properties to set the theme colours and fonts of the Mdm Explorer website.

# Additional context

Update the `ThemeService.loadTheme` method and replace the hard coded values with loaded api property values. Use the new `/explorer/theme` endpoint which will return a JSON response like this:

```json
{
    "count": 3,
    "items": [
        {
            "id": "f5a897f5-3e69-4c49-9795-827caacb27d9",
            "key": "material.colors.primary",
            "value": "#19381f"
        },
        {
            "id": "a83f3d68-c211-4c35-9bf9-83ffa91e328e",
            "key": "material.colors.accent",
            "value": "#cdb980"
        },
        {
            "id": "0eabbbe6-caee-4e4d-883c-84c70289849c",
            "key": "material.colors.warn",
            "value": "#a5122a"
        }
    ]
}
```

See MauroDataMapper-Plugins/mdm-plugin-explorer#18 for a table of properties that will returned, noting:

1. The `explorer.theme.` prefix will be removed from this list
2. The list returns a cutdown version of a list of API Property objects - only the key and value are really needed.
3. For typography values returned, they will look similar to "14px, 20px, 400". These are space separated values in this order: `fontSize`, `lineHeight`, `fontWeight`

Ensure these map to a `Theme` object correctly to work with the `ThemeService.applyTheme()` function.

pjmonks:
**Note:** this PR depends on MauroDataMapper-Plugins/mdm-plugin-explorer#19 being completed first. You can test this PR by building a local copy of `mdm-plugin-explorer` for now. This PR will be taken out of draft once the dependency is merged.

- If error occurs, use default theme
- Otherwise, merge settings from server with default to pass to applyTheme()
- Update tests for ThemeService

Resolves #274


src/app/shared/theme.service.ts
```diff
@@ -148,6 +150,31 @@ const defaultTheme: Theme = {
...
+  items: KeyValueIdentifier[],
+  key: string,
+  defaultValue: ThemeMaterialTypographyLevel
+) => {
+  const value = getKviValue(items, key, '');
+  if (value === '') {
+    return defaultValue;
+  }
+
+  const parts = value.split(', ');
```
greg-at-occ, line 163: Since these values will come from typed user inputs, I think we need at least to be able to handle spacing variations like "size,height,weight" or " size , height  , weight "

greg-at-occ (changes requested):

src/app/shared/theme.service.ts
pjmonks: Yes, you're correct. I've updated this parsing to remove all whitespace before splitting on commas, and updated the tests to catch both possible cases.

greg-at-occ (approved):

Issue #255, opened 2023-02-10
Creating a new request from the Request Details page does not refresh menu


pjmonks:
# Description

![image](https://user-images.githubusercontent.com/3219480/218060274-11e6ec45-34c6-4c15-b030-2790efe6ddb9.png)

The "Create request" button used in the Request Detail page to add elements to other requests will create a new request and map the data element to that new request, but the list of requests created does not refresh to show the new request.

# Steps to reproduce

1. View any data request detail, either unsent or submitted.
2. Click the "+" button next to a element, class or schema
3. Click the "Create new request" button in the popup menu
4. Enter a name for the request that has not been used before, then click "Create request"
5. Let the operation complete. The success dialog should appear and you should the continue browsing.
6. Click the same "+" button again and check the tick list. The new request you created is not there.
7. Do a hard refresh of the browser page and check the "+" menu again. The new request is now there and ticked.

# Expected behavior

After the operation has completed successfully, the menu should automatically refresh without the user having to do anything.

# Additional context

A broadcast event called `data-request-added` is sent globally via the `DataRequestsService` when the request is created and updated. I think this needs to be listened to in the Request Detail page to update the list of request intersections, which seems to be missing at the moment.

PR #301, opened 2023-03-02
Fixed checkmarks on dataelement and search.

OButlerOcc:
Fixed checkmarks not appearing in page when adding an element to a request.
Affects data element view and search/results view.

Possible continueation.
Implement broadcast state sync service to update intersections on data request view. search view and data element view to sync state across multiple tabs.

### DIFF
--- a/src/app/pages/data-element/data-element.component.ts
+++ b/src/app/pages/data-element/data-element.component.ts
@@ -28,7 +28,7 @@ import {
 } from '@maurodatamapper/mdm-resources';
 import { ProfileService } from 'src/app/mauro/profile.service';
 import { ToastrService } from 'ngx-toastr';
-import { switchMap, forkJoin, catchError, EMPTY, map } from 'rxjs';
+import { switchMap, forkJoin, catchError, EMPTY, map, takeUntil, Subject } from 'rxjs';
 import { BookmarkService } from 'src/app/data-explorer/bookmark.service';
 import { DataModelService } from 'src/app/mauro/data-model.service';
 import {
@@ -41,6 +41,7 @@ import {
   DataRequestsService,
 } from 'src/app/data-explorer/data-requests.service';
 import { TerminologyService } from 'src/app/mauro/terminology.service';
+import { BroadcastService } from 'src/app/core/broadcast.service';
 
 @Component({
   selector: 'mdm-data-element',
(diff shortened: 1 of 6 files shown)


src/app/pages/data-element/data-element.component.ts
```diff
@@ -247,4 +253,16 @@ export class DataElementComponent implements OnInit {
         .filter((section) => section.fields.length > 0),
     };
   }
+
+  private subscribeDataRequestChanges() {
+    this.broadcast
+      .on('data-request-added')
+      .pipe(takeUntil(this.unsubscribe$))
+      .subscribe(() => {
```
CristianGimenez, line 261: Should this be a map or switch map to avoid having nesting subscriptions?

src/app/pages/my-request-detail/my-request-detail.component.ts
```diff
@@ -537,18 +535,6 @@ export class MyRequestDetailComponent implements OnInit, OnDestroy {
     );
   }
 
-  private subscribeDataRequestChanges() {
```
CristianGimenez, line 540: This code was added in #281 which solves the issue #255. I've retested this, and it seems by removing this, we have reintroduced this bug, so we need to keep this code here.

CristianGimenez (changes requested):
Great work Oliver, for my testing you have solved the issue and the tick there looks great. However, I've make a couple of comments that you should address before approving this. Any questions, please let me know.

src/app/pages/my-request-detail/my-request-detail.component.ts
OButlerOcc: I have restored this code

src/app/pages/data-element/data-element.component.ts
OButlerOcc: Is that not only the case if we want to cancel the first subscription on the trigger? I think we want to stay subscribed so other events also trigger this method? Unless I'm being dense?

CristianGimenez: I believe your latest code solves this, thanks!

CristianGimenez (approved):
Thanks @OButlerOcc for addressing my comments, all looks good now, works as expected, the issue is solved. Great work!

PR #302, opened 2023-03-07
gh-154: Persistence of selection between search result pages

### LINKED ISSUE #154: Check boxes are not persisted in search results
### Issue
After Data Points have been searched for a list of valid results is displayed.  The user can tick the ones they want to add to their Request and then click on "With selected..." to add them. If the user clicks a couple of items and then clicks on the second page of results, then the page refreshes and the "With selected..." option is now disabled. Those selections have been lost.
### Solution
There are a couple of ways to solve this:
1. Prevent the user from moving to the second page if they have items ticked. Display a warning to the user saying that they have selected items and do they wish to add them to the request before moving on.
2. Remember the items that have been ticked on each page so that when the "With selected..." button is clicked they all get added. This is probably more complicated to achieve though.

greg-at-occ:
Introduces a new SelectionService that represent's the user's current selection of elements. 

Current limitations:

- The "current selection" element scrolls out of view for the lower half of the results list
- The selection persists regardless of user logins - it would carry from user to another if they log in on the same machine & browser.
- No "show all" page size available. The pageSize parameter already uses "undefined" to mean "use default" so it would need some rework to represent "no limit" as well.

Closes #154


src/app/data-explorer/selection-compact/selection-compact.component.ts
```diff
@@ -0,0 +1,61 @@
...
+    return this.selectionService.list$;
+  }
+
+  clearSelection() {
+    this.selectionService.clearSelection();
+  }
+
+  showSelection() {
+    if (this.sourceTargetIntersections !== undefined) {
+      this.matDialog.open<SelectionExpandedDialogComponent, SelectionExpandedDialogData>(
```
pjmonks, line 52: I've found a bug which seems to happen when I open this dialog, but I can't see an obvious cause for it at the moment.

1. Search for any term which will take you to the search listing page - note the URL will include query params like "?search=batch` etc
2. Tick a search result to add to the selection and make the "Current selection" box appear on the left.
3. Click on the "X element(s) selected" link

The `SelectionExpandedDialogComponent` opens correctly, but notice behind the dialog that the search listing page refreshes. It has somehow triggered a new search with no query params, effectively doing a search over every element in the catalogue.

I'm not sure why this is happening, please could you investigate.

src/app/data-explorer/selection-expanded-dialog/selection-expanded-dialog.component.html
```diff
@@ -0,0 +1,39 @@
...
+    <mat-checkbox checked="checked" (click)="deselectElement(element.id)"> </mat-checkbox>
+    <span>{{ fullPathFor(element) }}</span>
+  </div>
+</mat-dialog-content>
+<mat-dialog-actions align="end">
+  <mdm-data-element-multi-select
+    [sourceTargetIntersections]="sourceTargetIntersections"
+    [dataElements]="(selectedElements$ | async) ?? []"
+  ></mdm-data-element-multi-select>
+  <button mat-flat-button color="warn" (click)="clearSelection()">Clear selection</button>
```
pjmonks, line 38: See my previous comment about maybe using `mat-stroked-button` for the clear button.

src/app/data-explorer/selection-expanded-dialog/selection-expanded-dialog.component.html
```diff
@@ -0,0 +1,39 @@
...
+</div>
+<mat-dialog-content>
+  <p>{{ (selectedElements$ | async)?.length }} elements selected</p>
+  <div *ngFor="let element of selectedElements$ | async">
+    <mat-checkbox checked="checked" (click)="deselectElement(element.id)"> </mat-checkbox>
+    <span>{{ fullPathFor(element) }}</span>
+  </div>
+</mat-dialog-content>
+<mat-dialog-actions align="end">
+  <mdm-data-element-multi-select
```
pjmonks, line 34: I've noticed that the "Current selection" box has the Clear button then the "With Selected..." button, but in this dialog the order is swapped. I think it would be nicer if the button order was consistent, but I have no strong opinion on which way round both should be.

src/app/data-explorer/selection-expanded-dialog/selection-expanded-dialog.component.html
```diff
@@ -0,0 +1,39 @@
...
+-->
+
+<div mat-dialog-title class="dialog-title">
+  <h1>Selected data elements</h1>
+  <button mat-icon-button aria-label="close dialog" mat-dialog-close>
+    <mat-icon>close</mat-icon>
+  </button>
+</div>
+<mat-dialog-content>
+  <p>{{ (selectedElements$ | async)?.length }} elements selected</p>
```
pjmonks, line 27: I've noticed your use of the `async` pipe which looked interesting. Seeing as it is used in multiple places though, I wanted to check if it was being continually called. I updated the `SelectionService` to print this to the console to check:

```ts
public get list$(): Observable<DataElementSearchResult[]> {
    return this.listSubject
      .asObservable()
      .pipe(tap(() => console.log('Getting selection list...')));
  }
```

And this is the result:

image

This number shoots up as I move the mouse around the page too. I think what is happening is the Angular change detector keeps thinking these controls are changing, which happens on DOM event handlers. This is actually a pitfall of Angular change detection, because your templates are binding to a _function_ instead of a _value_ (even a `get` property is treated as a function), so Angular keeps calling the function to see if has changed. I would have though the `async` pipe avoids this because it is actually suggested by the Angular team themselves though.

To get around this issue though, I would suggest have your components call the `SelectionService.list$` property once, subscribe to it, then set the returned value to a local field in the component and bind to that in the template. This should prevent the change detector from going crazy.

src/app/data-explorer/selection-expanded-dialog/selection-expanded-dialog.component.ts
```diff
@@ -0,0 +1,69 @@
...
+  templateUrl: './selection-expanded-dialog.component.html',
+  styleUrls: ['./selection-expanded-dialog.component.scss'],
+})
+export class SelectionExpandedDialogComponent {
+  selectedElements$: Observable<DataElementSearchResult[]>;
+  sourceTargetIntersections: DataAccessRequestsSourceTargetIntersections;
+  constructor(
+    public dialogRef: MatDialogRef<SelectionExpandedDialogComponent>,
+    private selectionService: SelectionService,
+    @Inject(MAT_DIALOG_DATA) private data: SelectionExpandedDialogData
```
pjmonks, line 42: If `data` is not used in another class function then it doesn't need to be marked as a `private` field.

src/app/data-explorer/selection-expanded-dialog/selection-expanded-dialog.component.ts
```diff
@@ -0,0 +1,69 @@
...
+  clearSelection() {
+    this.selectionService.clearSelection();
+    this.dialogRef.close();
+  }
+
+  deselectElement(elementId: Uuid) {
+    this.selectionService.remove([elementId]);
+  }
+
+  fullPathFor(element: DataElementSearchResult) {
```
pjmonks, line 61: I thought we had a common function for working out the breadcrumb path, but it's actually a component to render called `mdm-breadcrumb`. You could potentially use that instead or keep this as it is.

src/app/data-explorer/selection-expanded-dialog/selection-expanded-dialog.component.html
pjmonks: I've just remembered something else if going down this route. Looking at your `SelectionService`, the `list$` is an observable that doesn't complete, so you will need to manually unsubscribe from it to avoid memory leaks. There are articles explaining this and a way of declaratively unsubscribing which you may find useful. Here is one:

https://medium.com/angular-in-depth/the-best-way-to-unsubscribe-rxjs-observable-in-the-angular-applications-d8f9aa42f6a0

src/app/data-explorer/selection.service.ts
```diff
@@ -0,0 +1,113 @@
...
+
+  private storeAndEmit() {
+    const selectionArray = Array.from(this.selected.values());
+
+    localStorage.setItem(this.STORAGE_KEY, JSON.stringify(selectionArray));
+
+    this.listSubject.next(selectionArray);
+  }
+
+  // On hold: Do we really need per-user storage? It's complicated
```
pjmonks, line 108: This is a clean and effective service 👍 I agree with your comment here, for now it doesn't need to be more complicated than this.

src/app/pages/search-listing/search-listing.component.html
```diff
@@ -57,6 +57,16 @@
               {{ resultSet?.totalResults }} results found
             </ng-container>
           </div>
+          <div>
```
pjmonks, line 60: Would it be possible to align this control closer to the "Sort by" control? My feeling is the sort and page size controls should act as a related group.

image

The inclusion of the page size control is welcome though.

pjmonks (changes requested):
Overall I like this feature change, I had my doubts about how easy it would be for a user to track what selections they were making over multiple pages, but this actually feels quite simple to use.

I have noted some areas of feedback and found a bug in using it, if these could be addressed please.

Also, one open question which can be debated: After creating a request from a selection using this system, should the selection then be cleared so it starts over again?

src/app/data-explorer/selection-compact/selection-compact.component.html
```diff
@@ -0,0 +1,49 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+SPDX-License-Identifier: Apache-2.0
+-->
+<div
+  class="highlight-box mdm-selection-compact"
+  *ngIf="(selectedElements$ | async)?.length != 0"
```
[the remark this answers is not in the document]
greg-at-occ: In this case I think it's correct because we may also have _undefined_ - we want the condition to be true if X != 0 AND X != undefined. It would be clearer to write the conditions separately but the cost would be to repeat the binding twice. I'm tempted to leave it but I don't like it much

src/app/data-explorer/selection-compact/selection-compact.component.html
```diff
@@ -0,0 +1,49 @@
...
+        aria-label="show selected elements"
+      >
+        {{ (selectedElements$ | async)?.length }} element(s) selected<mat-icon
+          fontSet="fa-solid"
+          fontIcon="fa-window-restore"
+        ></mat-icon>
+      </a>
+    </p>
+    <div class="mdm-selection-compact__footer">
+      <!-- button mat-flat-button color="accent" (click)="showSelection()">Open</button-->
```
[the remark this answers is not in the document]
greg-at-occ: Done

src/app/data-explorer/selection-compact/selection-compact.component.html
```diff
@@ -0,0 +1,49 @@
...
+      <button mat-flat-button color="warn" (click)="clearSelection()">Clear</button>
```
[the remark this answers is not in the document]
greg-at-occ: Agreed, done

src/app/data-explorer/selection-compact/selection-compact.component.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+
+      const spy = jest.spyOn(selectionService, 'clearSelection');
+
+      harness.component.clearSelection();
+
+      expect(selectedElements).toHaveLength(0);
+      expect(spy).toHaveBeenCalled();
+    });
+
+    // Open dialog
```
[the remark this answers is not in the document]
greg-at-occ: Left them in accidentally from outlining. Gone

src/app/data-explorer/selection-expanded-dialog/selection-expanded-dialog.component.html
greg-at-occ: I have fixed both of these bugs: the results of calling asObservable are now stored by the constructor and made available via a public member without a getter. I have replaced a getter with a public member in one of the other components too. Finally I added a `takeUntil(this.unsubscribe$)` to the subscription in search-listing.

src/app/data-explorer/selection-expanded-dialog/selection-expanded-dialog.component.ts
greg-at-occ: I found that component and decided not to use it - but looking again I can't see any good reason why. I must have misread it. Anyway I've replaced my hand-rolled function with this.

src/app/data-explorer/selection.service.ts
greg-at-occ: I've removed the commented-out code

src/app/pages/search-listing/search-listing.component.html
greg-at-occ: Agreed, done

src/app/data-explorer/selection-compact/selection-compact.component.html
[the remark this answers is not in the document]
pjmonks: OK, I didn't think of an `undefined` check. Acceptable to leave as-is then.

src/app/data-explorer/selection-compact/selection-compact.component.ts
pjmonks: I've re-tested and the odd refresh behaviour has now gone 👍

src/app/data-explorer/selection-expanded-dialog/selection-expanded-dialog.component.html
pjmonks: I see this has been updated now 👍

src/app/data-explorer/selection-expanded-dialog/selection-expanded-dialog.component.html
pjmonks: Re-tested usage of the `list$` observable by intercepting it with `tap(() => console.log())`, it now performs much better and only when actually required 👍

pjmonks (approved):
I've re-tested everything and am happy with the changes, especially with the `async` pipe usage and performance. Nice one @greg-at-occ !

Approved ✔️

Issue #276, opened 2023-02-17
Add Theming to Mdm Explorer: Load images from backend 


NigelPalmer:
Please fill in the sections below, adding/removing information where applicable:

# Description

After MauroDataMapper/mdm-core#394 has been completed, we need to load the header logo from the backend and display it on the Mdm Explorer pages.

# Additional context
The logo is currently loaded from 'assets/images/mauro-data-explorer.svg'. This will need to be changed to load an image from the database. Have a look at the Mdm-UI to see how the image is being loaded and displayed to the user in the Api Properties for setting up the theme. Or if this task has not been completed yet, have a look at the User profile image and how that is loaded and displayed.

PR #307, opened 2023-03-16
gh-276: Add Theming to Mdm Explorer: Load images from backend

NigelPalmer:
#276


src/app/shared/header/header.component.html
```diff
@@ -19,7 +19,10 @@
 <header class="mdm-header">
   <div class="mdm-header__logo">
     <a [routerLink]="logoLink?.routerLink ?? ''">
-      <img [src]="logoLink?.imageSrc" [alt]="logoLink?.label" />
+      <img
```
pjmonks, line 22: The custom logo functions well. I think there should be some size restrictions though to ensure the header is not growing too large. If someone is able to upload any  kind of image, they could upload one where the dimensions are unsuitable. For example:

image

Perhaps add a CSS class that imposes `max-width` and/or `max-height` in the logo region, but open to other suggestions.

src/app/shared/header/header.component.html
```diff
@@ -19,7 +19,10 @@
...
+        [src]="logoLink?.customImageSrc || logoLink?.defaultImageSrc"
```
pjmonks, line 23: By introducing a possible custom logo, we should ensure that the original "Mauro Data Mapper" logo is always guaranteed to appear somewhere on the page - in the `mdm-ui` is always appears in the footer for instance, so I think that `mdm-explorer` footer should be updated too. I'm not sure if that is in scope for this task or should be spun out to another task.

src/app/shared/theme.service.ts
```diff
@@ -232,6 +268,86 @@ export class ThemeService {
     this.applyMaterialPaletteToCss('warn', theme.material.colors.warn);
     this.applyTypographyToCss(theme.material.typography);
     this.applyAppColorsToCss(theme);
+    //this.getImageUrls(theme);
+  }
+
+  /*
+  getImageUrls(theme: Theme) {
```
pjmonks, line 275: Commented out code should be removed if it isn't required.

src/app/shared/theme.service.ts
```diff
@@ -186,14 +210,26 @@ interface Color {
   isLight: boolean;
 }
 
+export interface ThemeImageUrl {
+  url?: string;
+}
+
+let themeImageUrls: ThemeImageUrl[] = [];
```
pjmonks, line 217: This variable isn't referenced anywhere so could be removed.

src/app/shared/theme.service.ts
```diff
@@ -186,14 +210,26 @@ interface Color {
...
  * Service for managing theming of the application.
  */
 @Injectable({
   providedIn: 'root',
 })
 export class ThemeService {
-  constructor(private researchPlugin: ResearchPluginService) {}
+  constructor(
+    private researchPlugin: ResearchPluginService,
+    private endpoints: MdmEndpointsService,
```
pjmonks, line 228: `MdmEndpointsService` and `ToastrService` are not used in this service.

src/app/shared/theme.service.ts
```diff
@@ -232,6 +268,86 @@ export class ThemeService {
...
+            apiPropertyKey: apiProperty.key,
+            imageUrl: (apiProperty.value && isUUID(apiProperty.value))
+            ? `${environment.apiEndpoint}/admin/properties/${themeImage.apiPropertyId}/image`
+            : undefined;
+          } as ThemeImage;
+        });
+      });* /
+  }*/
+
+  private isUUID(value: string): boolean {
```
pjmonks, line 339: This feels like a utility function to me. Perhaps it could be moved as a function to `shared.types.ts` or somewhere similar?

pjmonks (changes requested):
Changing the logo to be customised "just works" ™️ 

Comments I've left focus on making sure the custom logo doesn't take up too much space and tidying up your temp code in `theme.service.ts`. Otherwise looking good so far 👍

pjmonks:
One other thing I've just thought of - the `docs/theming.md` file should also be updated to be correct based on theme image changes added. I've also spotted a big TODO comment about describing how loading themes work (my bad), could you possibly update that too at the same time please?

src/app/shared/header/header.component.html
NigelPalmer: My view is that an image should be correctly sized before it appears in the header. This is to be used for theming purposes, so I don't think an image should be being resized. But I guess this is open to abuse.

src/app/shared/header/header.component.html
NigelPalmer: I would prefer this to be addressed in a follow-up task.

src/app/shared/header/header.component.html
pjmonks: Based on your opinion, I would at least suggest the following short-term edits:

1. Document what maximum image size is expected to at least have it recorded somewhere, then assume that the end user will adhere to that.
2. (Optional) Consider the layout of the rest of the header with different sizes of logos. In the above example, the links could be aligned vertical centre for instance. Happy for that to be a separate bug issue to resolve later though.

src/app/shared/header/header.component.html
pjmonks: Created #310

src/app/shared/header/header.component.html
pjmonks: Updated `docs/theming.md` to define maximum logo size.

src/app/shared/theme.service.ts
pjmonks: Done

src/app/shared/theme.service.ts
pjmonks: Done

src/app/shared/theme.service.ts
pjmonks: Moved this function

pjmonks (approved):
Implemented feedback changes on behalf of @NigelPalmer

PR #331, opened 2023-05-03
gh-330: Create SDE Adaptor

### LINKED ISSUE #330: Create SDE Adaptor
# Description

To facilitate development of the web pages that interact with the SDE, whilst the backend is still being developed, an adaptor is required. This will live in the MDM-Explorer code base. The adaptor will have functions that will be called by the UI code which will interact with the backend to extract and post data. These functions will be defined in an interface. This will then allow us to have two implementations, one that talks to the real backend, and one that talks to a mock backend. The page developments can begin using mocks, and the two ends can be plumbed together later.

The adaptor in the MDM Explorer should only implement interactions that are required by the MDM Explorer UI. The SDE will have it's own adaptor to talk to the SDE backend. There may be duplication of some functionality in the two adaptors, but we are happy to live with this for now. If the duplication grows, then we will create a shared adaptor library that both the SDE Admin and MDE can use. But this is outside the scope of this task.

The functions in the Adaptor will return strongly typed objects.

The adaptor will need to be extended as pages are developed, since it will not be clear what is exactly needed until each page is worked on. As an initial implementation the adaptor should provide an interface to:

Load an Admin User
Load a Research User
Load an Organisation

Tests should be written to prove that the adaptor works against a mock implementation of the backend.

NigelPalmer:
This is an adaptor layer to fulfill #330 . The code has been organised so that all the fake code is in a folder together so that it can easily be removed down the line. Both real and fake rest handlers can live side by side, so that it will be possible to change where data is coming from iteratively, without having to do an all or nothing approach.

The tests written are to prove the fakes work, rather than to test the adaptor itself. Full tests will be required if this approach is accepted. 

The code written here will need to be copied to the SDE Admin UI. It became apparent whilst writing this, that a shared project would be required since a lot of this functionality will be shared between the two applications.

### DIFF
--- a/src/app/secure-data-environment/endpoints/organisation.endpoints.ts
+++ b/src/app/secure-data-environment/endpoints/organisation.endpoints.ts
@@ -35,12 +35,6 @@ export class OrganisationEndpoints {
     );
   }
 
-  getOrganisationMember(organisationMemberId: Uuid): Observable<OrganisationMember[]> {
-    return this.sdeRestHandler.get<OrganisationMember>(
-      `${SdeApiEndPoints.OrganisationMemberGet}${organisationMemberId}`
-    );
-  }
-
   listOrganisationMembers(organisationId: Uuid): Observable<OrganisationMember[]> {
     return this.sdeRestHandler.get<OrganisationMember[]>(
       `${SdeApiEndPoints.OrganisationMemberList}${organisationId}`
--- a/src/app/secure-data-environment/endpoints/organisation.endpoints.ts
+++ b/src/app/secure-data-environment/endpoints/organisation.endpoints.ts
@@ -1,0 +1,49 @@
+/*
+Copyright 2022-2023 University of Oxford
+and Health and Social Care Information Centre, also known as NHS Digital
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+import { Organisation, OrganisationMember } from '../resources/organisation.resources';
+import { Observable } from 'rxjs';
+import { ISdeRestHandler } from '../sde-rest-handler.interface';
+import { Uuid } from '@maurodatamapper/mdm-resources';
+import { SdeApiEndPoints } from './endpoints.dictionary';
+
+export class OrganisationEndpoints {
+  constructor(private sdeRestHandler: ISdeRestHandler) {}
+
+  listOrganisations(): Observable<Organisation[]> {
+    return this.sdeRestHandler.get<Organisation[]>(SdeApiEndPoints.OrganisationList);
+  }
+
+  getOrganisation(organisationId: Uuid): Observable<Organisation> {
+    return this.sdeRestHandler.get<Organisation>(
+      `${SdeApiEndPoints.OrganisationGet}${organisationId}`
+    );
+  }
+
+  getOrganisationMember(organisationMemberId: Uuid): Observable<OrganisationMember[]> {
+    return this.sdeRestHandler.get<OrganisationMember>(
+      `${SdeApiEndPoints.OrganisationMemberGet}${organisationMemberId}`
+    );
+  }
+
+  listOrganisationMembers(organisationId: Uuid): Observable<OrganisationMember[]> {
+    return this.sdeRestHandler.get<OrganisationMember[]>(
+      `${SdeApiEndPoints.OrganisationMemberList}${organisationId}`
+    );
+  }
+}
--- a/src/app/secure-data-environment/endpoints/user.endpoints.ts
+++ b/src/app/secure-data-environment/endpoints/user.endpoints.ts
@@ -1,0 +1,37 @@
+/*
+Copyright 2022-2023 University of Oxford
+and Health and Social Care Information Centre, also known as NHS Digital
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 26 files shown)


src/environments/environment.ts
```diff
@@ -23,6 +23,7 @@ SPDX-License-Identifier: Apache-2.0
 export const environment = {
   production: false,
   apiEndpoint: 'http://localhost:8080/api',
```
greg-at-occ: If this is the explorer plugin we probably want to call it mauroCoreEndpoint

src/app/secure-data-environment/endpoints/organisation.endpoints.ts
```diff
@@ -0,0 +1,49 @@
...
+    return this.sdeRestHandler.get<Organisation[]>(SdeApiEndPoints.OrganisationList);
+  }
+
+  getOrganisation(organisationId: Uuid): Observable<Organisation> {
+    return this.sdeRestHandler.get<Organisation>(
+      `${SdeApiEndPoints.OrganisationGet}${organisationId}`
+    );
+  }
+
+  getOrganisationMember(organisationMemberId: Uuid): Observable<OrganisationMember[]> {
```
greg-at-occ: We probably don't need this one - the "id" of the membership is never really used anywhere else

src/app/secure-data-environment/endpoints/user.endpoints.ts
```diff
@@ -0,0 +1,37 @@
...
+import { Observable } from 'rxjs';
+import { ISdeRestHandler } from '../sde-rest-handler.interface';
+import { Uuid } from '@maurodatamapper/mdm-resources';
+import { SdeApiEndPoints } from './endpoints.dictionary';
+import { AdminUser, ResearchUser } from '../resources/users.resources';
+
+export class UserEndpoints {
+  constructor(private sdeRestHandler: ISdeRestHandler) {}
+
+  getAdminUser(userId: Uuid): Observable<AdminUser> {
```
greg-at-occ, line 28: Research users won't need this but we could keep it for the admin side for now

src/app/secure-data-environment/endpoints/organisation.endpoints.ts
```diff
@@ -0,0 +1,49 @@
...
+
+SPDX-License-Identifier: Apache-2.0
+*/
+import { Organisation, OrganisationMember } from '../resources/organisation.resources';
+import { Observable } from 'rxjs';
+import { ISdeRestHandler } from '../sde-rest-handler.interface';
+import { Uuid } from '@maurodatamapper/mdm-resources';
+import { SdeApiEndPoints } from './endpoints.dictionary';
+
+export class OrganisationEndpoints {
```
greg-at-occ, line 25: Do we want one example of a command endpoint that performs a POST and creates/updates an entity?

src/app/secure-data-environment/resources/organisation.resources.ts
```diff
@@ -0,0 +1,39 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+import { Uuid } from '@maurodatamapper/mdm-resources';
+
+export interface Organisation {
+  id: Uuid;
+  createdAt: Date;
```
greg-at-occ, line 23: No action for now, but we will probably need to think about Date representation in JSON DTOs because it definitely won't be a JavaScript Date

src/app/secure-data-environment/sde-rest-handler.ts
```diff
@@ -0,0 +1,48 @@
...
+
+@Injectable({
+  providedIn: 'root',
+})
+export class SdeRestHandler implements ISdeRestHandler {
+  constructor(private httpClient: HttpClient) {}
+
+  get<T>(url: string): any {
+    return this.httpClient
+      .get<T>(url)
```
greg-at-occ, line 32: We may need to think about some error codes that contain useful info for end users e.g. 422 will often say what problems prevented the user's action - but that won't be for "get"

src/app/testing/testing.helpers.ts
```diff
@@ -79,6 +79,7 @@ export const setupTestModuleForService = <T>(
   TestBed.configureTestingModule({
     imports: [TestingModule, ...(configuration?.imports ?? [])],
     providers: configuration?.providers ?? [],
+    teardown: { destroyAfterEach: false },
```
greg-at-occ: Not sure what this does but it seems fairly broadly scoped - it won't break anything?

greg-at-occ (changes requested):
I think this looks like a solid approach that will decouple our UI work from our backend work as intended. A few comments and queries included.

src/environments/environment.ts
NigelPalmer: I have changed as many references as I sensibly can. Note, that some apiEndpoint variables are coming from mdm-resources. I don't want to change the name in there since that will then impact on the mdm-ui as well as mdm-explorer, which is too impactful a change.

src/app/secure-data-environment/endpoints/organisation.endpoints.ts
NigelPalmer: Deleted.

src/app/secure-data-environment/endpoints/user.endpoints.ts
NigelPalmer: I agree. I will leave it here for now to make it easier to copy this stuff to the SDE-Admin-UI. Then I will remove it.

src/app/secure-data-environment/endpoints/organisation.endpoints.ts
NigelPalmer: The mdm-explorer will not be posting any organisation related data. We will need to add this example in to the SDE-Admin-UI version of this task. It is scoped to do posts. See: https://github.com/MauroDataMapper/sde-admin-ui/issues/12

src/app/secure-data-environment/sde-rest-handler.ts
NigelPalmer: Again, we can address this when implementing some post commands in the sde-admin-ui.

src/app/testing/testing.helpers.ts
NigelPalmer: Deleted. This snuck in a part of trying to get the tests to prove a live httpClient request would work. This is not required for the committed code.

src/app/secure-data-environment/resources/organisation.resources.ts
greg-at-occ: Having looked at this, we should use string, namely the format used by JavaScripts Date.toJSON() = '2012-04-23T18:25:43.511Z'

greg-at-occ (approved):
Ok this looks good!

PR #332, opened 2023-05-12
Data specification versioning

### LINKED ISSUE #315: Create new versions of Data Specifications
# Description

Data Specifications, formerly known as Data Requests need to be versioned. When a data request is submitted the version number should be set. After it has been submitted, the user should be able to create a new version starting from the submitted one. We need to take advantage of Mauro's existing versioning functionality to do this.

In the Mdm-UI a user can finalise a data model (click on the green tick), at which point they get to choose the version number they wish to use. We need to use this same functionality in mdm-explorer, since a Data Specification is just a Data Model. We won't give the mdm-explorer user a choice though, mdm-explorer will simply set the "Major Version" number behind the scenes.

Once a data model is finalised in the Mdm-UI a user can create a new version by clicking on the "create a new version" button (2 docs in a pile, top right). This then asks the user to choose the type of version they wish to create, either fork, version or branch. For the Mdm Explorer we simply want to automatically select 'version' behind the scenes. The user should not be presented with a choice. Add a new button to the Data Specifications page that is only visible after a Data Specification has been submitted. When clicked it will create a new version.

If there are multiple versions with the same name, then only the most recent version should be displayed in the list of Data Specifications.

CristianGimenez:
As mentioned in the ticket, we have agreed not to implement the finalisation feature mentioned origianally in the issue. In this implementation you will find the following features

- Creating a new version of a finalised (submitted) data specification.
- When listing your data specifications, or shared data specifications now, only the latest model version is returned.
- On the details of a data specification there is a dropdown that can be used to see a different version.

Since there is no way to finalise the data specifications, an instance of mdm-ui is required to run manual tests, so we need to finalise manually data specifications to see the "create new version" button and to see how they are now listed or not.

Also the version selector component infers the current "draft" version to be the next available major version.

Draft until: [mdm-resources/106](https://github.com/MauroDataMapper/mdm-resources/pull/107) and [plugin-explorer/34](https://github.com/MauroDataMapper-Plugins/mdm-plugin-explorer/issues/34) are resolved

Closes #315


src/app/mauro/data-model.service.ts
```diff
@@ -368,7 +370,7 @@ export class DataModelService {
     }
 
     return this.endpoints.dataModel
-      .newBranchModelVersion(model.id, {})
+      .newBranchModelVersion(model.id, { asynchronous: false })
```
NigelPalmer: Why have we turned off asynchronus here?

src/styles/_versioncontrol.scss
```diff
@@ -0,0 +1,32 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+.version_control {
+  .select-label {
+    color: var(--theme-color-primary-contrast) !important;
```
NigelPalmer: This should be `$color-primary-contrast` . You should not be calling var(..) directly.

src/styles/_versioncontrol.scss
```diff
@@ -0,0 +1,32 @@
...
+  }
+  .mat-form-field-outline {
+    color: var(--theme-color-primary-contrast) !important;
```
NigelPalmer: This should be $color-primary-contrast . You should not be calling var(..) directly.

src/styles/_versioncontrol.scss
```diff
@@ -0,0 +1,32 @@
...
+  }
+  .mat-select-value {
+    color: var(--theme-color-primary-contrast) !important;
```
NigelPalmer: This should be $color-primary-contrast . You should not be calling var(..) directly.

src/styles/_versioncontrol.scss
```diff
@@ -0,0 +1,32 @@
...
+  }
+  .mat-select-arrow {
+    color: var(--theme-color-primary-contrast) !important;
```
NigelPalmer: This should be $color-primary-contrast . You should not be calling var(..) directly.

NigelPalmer (changes requested):
Overall good job, and the main functionality is there. We do have some problems though, that need to be fixed.

- [ ] I don't like the number 1 just floating on the header bar of a Data Specification. It has no context or meaning to the user and simple looks odd. Once submitted the version number changed to V1.0.0. It should be displayed like this from the beginning.  **- Please fix.**
- [X] Copied a specification and then deletes an element from the copy. The original kept that element. _No work required_
- [ ] I added a cohort query and data query to V2.0.0. When I change the version to be V1.0.0 the queries were still displayed. The queries should be associated with the version of the specification. (Or possibly this is the next couple of errors manifesting, read on...) **- Please fix.**
- [ ] Once I had V1.0.0, V2.0.0 and unsubmitted v3.0.0, if I changed the Cohort query for v3.0.0, then the same changes were applied to v1.0.0 (but NOT v2.0.0). Changing the Cohort Query should only affect the latest version. **- Please fix.**
- [ ] Same happened if I updated and saved the latest Data Query. The changes were also applied to V1.0.0. **- Please fix.**
- [ ] You can click "New Version" at any time. If you already have an unsubmitted version then you will get the error `Http failure response for http://localhost:8080/api/dataModels/9a93d3f1-ddd8-48e0-895a-6b577f0d3f5e/newBranchModelVersion: 403 OK`. The "New Version" option should only be available if ALL versions are submitted. I suggest disabling the button rather than hiding it. Once I get this error, I am unable to create new versions for any Data Specification until I log out and back in again. **- Please fix.**

CristianGimenez:
> Overall good job, and the main functionality is there. We do have some problems though, that need to be fixed.
> 
> * [ ]  I don't like the number 1 just floating on the header bar of a Data Specification. It has no context or meaning to the user and simple looks odd. Once submitted the version number changed to V1.0.0. It should be displayed like this from the beginning.  **- Please fix.**
> * [x]  Copied a specification and then deletes an element from the copy. The original kept that element. _No work required_
> * [ ]  I added a cohort query and data query to V2.0.0. When I change the version to be V1.0.0 the queries were still displayed. The queries should be associated with the version of the specification. (Or possibly this is the next couple of errors manifesting, read on...) **- Please fix.**
> * [ ]  Once I had V1.0.0, V2.0.0 and unsubmitted v3.0.0, if I changed the Cohort query for v3.0.0, then the same changes were applied to v1.0.0 (but NOT v2.0.0). Changing the Cohort Query should only affect the latest version. **- Please fix.**
> * [ ]  Same happened if I updated and saved the latest Data Query. The changes were also applied to V1.0.0. **- Please fix.**
> * [ ]  You can click "New Version" at any time. If you already have an unsubmitted version then you will get the error `Http failure response for http://localhost:8080/api/dataModels/9a93d3f1-ddd8-48e0-895a-6b577f0d3f5e/newBranchModelVersion: 403 OK`. The "New Version" option should only be available if ALL versions are submitted. I suggest disabling the button rather than hiding it. Once I get this error, I am unable to create new versions for any Data Specification until I log out and back in again. **- Please fix.**

I'm letting you check or uncheck those elements after you retest. I have just pushed a change fixing all the issues mentioned there but the ones about cohort and data queries which I'm investigating now.
I dont like how the disabled button looks like (and I think that is not a good UX practice to disable buttons without a clear indication of why are they disabled) but take a look and let me know what would you prefer. The button will be hidden if the dataspec is not submitted still.

src/styles/_versioncontrol.scss
CristianGimenez: done

src/styles/_versioncontrol.scss
CristianGimenez: done

src/styles/_versioncontrol.scss
CristianGimenez: done

src/styles/_versioncontrol.scss
CristianGimenez: done

src/app/mauro/data-model.service.ts
CristianGimenez: To be aligned with mdm-ui, changed back

CristianGimenez:
@NigelPalmer I have fixed all your the issues you raised and merge conflicts, so if you are happy to retest, this would be ready to merge now.

NigelPalmer (changes requested):
- [X] V1.0.0 now displayed in all cases on the Data Specification list.
- [X] Adding Cohort and Data queries only adds them to the latest version.

- [ ] The disabled button is so dark it is barely visible. We either need to disable it with a brighter colour or hide it.
- [ ] After I submit a specification, the "New Version" button does not activate until you click away and come back.
- [ ] When I try to create a new version after submitting I get a 403 error. Steps to reproduce:
1. Start a clean browser and go to MDE 
2. Log in.
3. Select a data specification and submit.
4. Click "Back to my Data Specifications"
5. Click on the same data specification
6. Click on "New Version"
7. Click on "Yes"
I can only clear this problem by closing the browser and reopening it. This makes testing a burden since only submitted Data Specifications can have new versions created so I am having to constantly close and open my browser to be able to test this work. (Found another way around, if I submit and then share with the community, then I am able to click back and then select the Data Specification again. Now when I click "New Version" no errors occur.)
- [ ] The version drop down is showing the selected version twice. (Or possibly it's a symptom of the next issue.)
DuplicatedVersionNumber
- [ ] I had 4 versions of the same Data Specification. I selected V2 and clicked on "New Version". This created a new v3, so now I have two v3s.
CreatedNewVersionPlusOne
- [ ] I created a brand new Data Specification. This resulted in two version 1s.
DuplicatedOnes

CristianGimenez:
@NigelPalmer I have pushed a new commit with the following fixes/changes:
- New treeList method in the folders endpoint so we can replicate the "refresh" of mdm-ui, solving the submit problems.
- Refactored my-data-specification-detail.component.ts to remove the nested subscriptions that I've found.
- Changed the styles of the new version button, hopefully is more visible now, I've also added a tooltip (which requires a bit more of boiler plate since it does not work well in disabled buttons)
- Fixed bugs mentioned as duplicated versions etc.

Hopefully this is now working well and can be merged, could you retest and let me know any further changes required?

src/styles.scss
```diff
@@ -37,3 +37,4 @@ Import more theme SASS files here to make them available to the app
 @import "styles/home";
 @import "styles/querybuilder";
 @import "styles/versioncontrol";
+@import "styles/_dataspecificationsactions";
```
NigelPalmer: I don't think the proceeding _ should be here since it is not present on any of the other references.

src/app/pages/my-data-specification-detail/my-data-specification-detail.component.ts
```diff
@@ -203,12 +203,13 @@ export class MyDataSpecificationDetailComponent implements OnInit, OnDestroy {
       )
       .subscribe((dataModel) => {
         // Refresh the current state of the data specification in view
-        this.dataSpecification = mapToDataSpecification(dataModel);
+        this.setDataSpecification(mapToDataSpecification(dataModel));
+        // this.dataSpecification = mapToDataSpecification(dataModel);
```
NigelPalmer: Comment should be deleted

NigelPalmer (changes requested):
- [X] New Version button is now working correctly.
- [X] No errors are occuring.
- [ ] Tests are still failing in Jenkins?
- [ ] Please hide the "New Version" button instead of disabling it when it is not required.  Apologies for messing you about with this, but having tested it on this iteration, it has become clear to me that hiding the button makes a better user experience.

src/app/pages/my-data-specification-detail/my-data-specification-detail.component.ts
CristianGimenez: deleted in the latest version

src/styles.scss
CristianGimenez: You are right, alas I'm deleting that file anyway since it is no longer needed (it was all about the disabled button styles)

NigelPalmer (approved):
Button now removed. I am not completely convinced by the version dropdown box jumping about, but I think it makes for a better user experience not having the button visible when it is not required. We can review this when creating the request management pages. Approved.

PR #340, opened 2023-10-02
gh 339 - sde login

### LINKED ISSUE #339: SDE Login
# Description

The MDE needs to be able to sign into the SDE Researcher API. For now this can be done with the impersonate end point.

# Additional context

SDE endpoints in the MDE are defined in `app > secure-data-environment > endpoints`. Add a new entry to the SdeApiEndPoints table that points at `/impersonate/with-token`.

Update `sde-rest-handler` so that it has put and post functionality. This will not be needed for logging in, but will be needed soon, so may as well add it in.

When a user signs in to the mdm-explorer we then need to sign them in to the SDE. This should be done once the mdm-explorer login has been accepted. The rest handler will not be able to be used to send the sign-in request because it will not have the token that is required to make the call. Have a look at `AuthenticationService` in the `sde-admin-ui` to see how this is done. The users email address will need to be passed in as a payload. The most important part is that the token returned must be stored somewhere and then appended to requests by the `sde-request-handler`. Again look in the `sde-admin-ui` for ideas as to how this could be done. Another thing to be wary of is that a user could be registered with the MDE and not the SDE. This systems should be aligned, but it is possible they won't be. So make sure that any failure to login to the SDE is handled gracefully. I think we should still allow access to the MDE, but disable the SDE tab if the login for the SDE has failed.

Don't worry about password handling for now. We will come back to that in a later task.

abwilson23:
Attempts to log user into the sde discreetly upon logging in to the mde. The token returned is appended to the headers of subsequent requests made to the sde api.


NigelPalmer (dismissed):
This is looking good. It is clear whether a user has automatically logged into the SDE or not. The tidying up of the code base is appreciated as well.

NigelPalmer (approved):

PR #344, opened 2023-10-09
gh-343: Fix SDE environment settings for prod

### LINKED ISSUE #343: SDE: Fix production environment settings.
Please fill in the sections below, adding/removing information where applicable:

# Description

The new sde endpoint environment settings were missed from the production configuration, which is preventing the production build from completing successfully. The sde end point needs to be copied into the production build.

NigelPalmer:
The sdeResearcherEndpoint environment variable has been added to the production build. Also note that sdeAdminEndpoint has been removed since this should not be present in mdm-explorer.


OButlerOcc (approved):
looks simple enough to me

PR #345, opened 2023-10-30
gh-342 Sign in to SDE as a researcher using OpenID Connect

### LINKED ISSUE #342: SDE: Login using Azure AD
# Description

When using a login configured with OpenID to login to Mdm Explorer the research user should also be logged into the SDE (secure Data Environemnt). This will probable involve having to pass a token to the SDE backend.

# Additional context

MauroDataMapper/sde-core#22 has been created to implement this functionality in the backend.

pjmonks:
Resolves #342 

**Note:** this PR must be reviewed alongside:

- MauroDataMapper/sde-core#114
- MauroDataMapper/sde-resources#75

These PRs should be accepted and merged first.

- Fix OpenID Connect authorization for Mauro sign in
  - Correctly redirects to dashboard page and updates header state with signed in user
  - Remove fake SDE sign in call, not required
- Create SDE module for components
- Adjust existing SDE main page to use new components
- Create sign in component for SDE researcher. Lists all available OIDC providers from SDE to trigger redirect and authentication
- Add redirect page for SDE authentication to complete login process
  - Micronaut backend will redirect to this page to complete SDE login
  - Successful login from OIDC will fetch user details of logged in user and put them in local storage
  - Redirect to correct page and display user details
  - Update SdeRestHandler to use `withCredentials` option to always send authorization cookies back to SDE backend
- Sign in is automatic when Mauro SSO is used, with the proviso that:
  - Mauro and SDE OpenID Connect providers are configured exactly the same.
  - Providers share the same labels (to synchronise them)
  - Then SDE sign in occurs automatically as soon as Mauro sign in succeeds
 - Sign out is automatic when Mauro SSO is used:
   - A hidden HTML form had to be put on the `app.component`, which is where the original Mauro sign out event is handled
   - Once Mauro is signed out, the hidden HTML form is automatically submitted to trigger a HTTP POST request to the SDE

Notes about testing:

1. OCC logins via Azure AD mean we have two accounts available for SSO - OCC and SystemC. Because of the way I had to implement the SSO login loop - login to Mauro session first, then automatically login to SDE session - you will see the Azure account selection page twice to select an account (pick OCC both times). This duplication is not desirable but needs to be addressed at a later date.


src/app/app.component.ts
```diff
@@ -283,19 +297,13 @@ export class AppComponent implements OnInit, OnDestroy {
...
-        this.userDetails.set(user);
-
-        return user;
-      })
-    );
+      .subscribe(() => {
+        if (this.sdeSignOutForm) {
+          // If SDE sign out form is available, trigger a form POST request
+          // This will force the app to sign out of the SDE too automatically
+          this.sdeSignOutForm.nativeElement.submit();
```
greg-at-occ, line 315: Is it not possible to do this sign out with a direct XHR call? I'm curious why we have this hidden form approach.

src/app/app.component.ts
```diff
@@ -184,7 +198,8 @@ export class AppComponent implements OnInit, OnDestroy {
     private toastr: ToastrService,
     private userIdle: UserIdleService,
     private error: ErrorService,
-    private themes: ThemeService
+    private themes: ThemeService,
+    private sdeAuthentication: AuthenticationEndpoints
```
greg-at-occ: I think we will want to move this into sde-resources

src/app/secure-data-environment/components/organisation-list/authentication.resources.ts
```diff
@@ -0,0 +1,53 @@
...
+*/
+import { Uuid } from '@maurodatamapper/mdm-resources';
+
+export interface SdeOpenIdConnectProvider {
+  name: string;
+  label: string;
+  imageUrl?: string;
+}
+
+export const sdeOpenIdConnectProviders: SdeOpenIdConnectProvider[] = [
```
greg-at-occ: This is in sde-resources now, should it be removed here?

src/app/secure-data-environment/resources/authentication.resources.ts
```diff
@@ -0,0 +1,47 @@
+/*
```
greg-at-occ: There are two copies of authentication.resources.ts in different places here, plus another in sde-resources. Can we dispose of some of them?

src/app/app.component.ts
```diff
@@ -42,14 +48,20 @@ import { FooterLink } from './shared/footer/footer.component';
...
   selector: 'mdm-root',
   templateUrl: './app.component.html',
   styleUrls: ['./app.component.scss'],
 })
 export class AppComponent implements OnInit, OnDestroy {
+  /**
+   * Hidden HTML form element to manually trigger a sign out of the SDE.
+   * This is automatically handled in the broadcast 'sign-out-user' event.
+   */
+  @ViewChild('sdeSignOutForm') sdeSignOutForm?: ElementRef<HTMLFormElement>;
```
greg-at-occ, line 69: I think I commented on this before but I can't find my comment right now: do we need a hidden form for this? Is there not an XHR approach?

src/app/pages/sign-in/sign-in.component.ts
```diff
@@ -80,10 +82,29 @@ export class SignInComponent implements OnInit {
...
-    localStorage.setItem('openIdConnectProviderId', provider.id);
-    const redirectUrl = this.security.getOpenIdConnectAuthorizationUrl(provider);
-    window.open(redirectUrl.toString(), '_self');
+    // Try to synchronise the Mauro and SDE OpenID Connect providers to sequence in order
+    this.authenticationEndpoints
+      .listOauthProviders()
+      .pipe(
+        catchError(() => []),
+        map((sdeProviders) =>
+          // Big assumption - that Mauro and SDE have OpenID Connect providers are configured
```
greg-at-occ, line 91: Can we use a more reliable detail than the label like the URI for the IDP?

src/app/secure-data-environment/sde-sign-in/sde-sign-in.component.html
```diff
@@ -0,0 +1,62 @@
+<!--
```
greg-at-occ, line 1: Possibly combine with equivalent component in sde-admin-ui?

src/app/security/user-details.service.ts
```diff
@@ -105,4 +106,37 @@ export class UserDetailsService {
...
+    const userEmail = localStorage.getItem('sdeEmail');
+    if (!userEmail || userEmail.length === 0) {
+      return null;
+    }
+
+    return {
+      id: localStorage.getItem('sdeUserId') ?? '',
+      email: localStorage.getItem('sdeEmail') ?? '',
+      preferredName: localStorage.getItem('sdePreferredName') ?? '',
+      isDeleted: false,
```
greg-at-occ: Is filling in all these fields with default values a sign of a problem elsewhere?

greg-at-occ (changes requested):
A bit of tidy up, and could you confirm that you decided against combining a couple of components into sde-resources?

src/app/secure-data-environment/sde-sign-in/sde-sign-in.component.html
pjmonks: I've now changed this to use a shared component from `sde-resources`

src/app/app.component.ts
pjmonks: Unfortunately not. I've updated this code with a more detailed comment explaining this bizarre workaround.

src/app/app.component.ts
pjmonks: Unfortunately not. I've updated this code with a more detailed comment explaining this bizarre workaround.

src/app/app.component.ts
pjmonks: This has been done now.

src/app/secure-data-environment/components/organisation-list/authentication.resources.ts
pjmonks: Yes, I've removed this obsolete copy.

src/app/secure-data-environment/resources/authentication.resources.ts
pjmonks: Yes, I've removed this obsolete copy.

src/app/pages/sign-in/sign-in.component.ts
pjmonks: Unfortunately I don't think we can at the moment, it's a problem with storing the same Oauth provider details in two different systems. I've elaborated more in the comment instead.

src/app/security/user-details.service.ts
pjmonks: This function wasn't strictly necessary so I've removed it instead.

src/app/pages/sign-in/sign-in.component.ts
greg-at-occ: In other words we would need to make changes to the MDM core OIDC plugin to support a more robust approach?

greg-at-occ (approved):
This is great and works so smoothly

PR #355, opened 2023-11-17
gh-354-use-workflow-resolver

### LINKED ISSUE #354: Load workflow data via resolver
Use the newly added WorkflowResolver to retrieve and cache workflow/requestStatusOptions data in the mdm.

abwilson23:
Add resolver to the SDE route. 

To test: verify that the workflow is loaded only upon navigating to the SDE route. 

closes #354


NigelPalmer (approved):
This is now working in a good way with the request statuses being successfully loaded from the sde-core.

Issue #351, opened 2023-11-13
Complete Requests functionality (Epic)


NigelPalmer:
This issue is to group all the work that is related to the request page so that we can make sure it is merged at the same time.

# Allow researcher to create requests

- [x] maurodatamapper/sde-core#128
- [x] maurodatamapper/sde-resources#34
- [x] #347 

Note the changes implemented in the PRs for the above issues focus on these agreed requirements (slightly different to #347):

1. Allow a researcher to create a new request in "draft" status
2. Submit that request to "Awaiting approval" status
3. Allow a researcher to perform request actions for submitting a request approval

No other request status transitions are supported for a researcher yet.

Also, the list of requests a researcher sees is complete from a front-end perspective, but requires more backend work - see MauroDataMapper/sde-core#132.

# Additional changes

These changes focus on making sure other parts of the SDE are still functioning after altering the shared backend and component library.

- [x] maurodatamapper/sde-admin-ui#127
- [x] maurodatamapper/sde-resources#37

NigelPalmer:
All issues in this epic have been merged.

PR #356, opened 2023-11-23
gh-347 Update to display correct requests list for researchers

### LINKED ISSUE #347: SDE: Request details page (Full functionality)
# Description

#326 integrated the `sde-resources` library into `mdm-explorer`. The Requests list component was added and this functions, but it does not do everything it should. This task is to get the request list and workflow behaving as it should for a researcher.

1. The Researcher needs to be able to create a new request. This should be done using the existing dialog for showing requests if possible.
2. The Researcher needs to be able to submit a draft request. The workflow of what action buttons are displayed when, is working correctly as based on the workflow, but when they are clicked errors occur. Changes in the `sde-core` will be required to support this.
3. The list component has not been properly configured. It should be setup to show the Researchers own requests and no other requests. The filter should not be an available option. This should be a simple case of configuring the list component in the correct way. See the `sde-admin-ui` for examples of list component configuration.
4. Researchers should be able to edit Requests when they have action buttons displayed, but otherwise the edit button should not be available. This should just work, but needs checking.

Supported by maurodatamapper/sde-resources#34
Supported by maurodatamapper/sde-core#128

pjmonks:
Resolves #347 

- Supports the epic #351 
- Depends on MauroDataMapper/sde-core#128
- Depends on MauroDataMapper/sde-resources#34

Small updates to have `mdm-explorer` show the correct request list for researchers - most of the functionality for handling creation of requests is handled in the shared library.

# Notes

- Set correct display mode for `sde-requests-list` component

# Additional

- Update `angular.json` to support debugging of the `sde-resources` library/node package

# Testing

This PR, along with its dependencies, should support:

1. A researcher being able to create requests - into "Draft" status
2. Then being able to submit that request to "Awaiting Approval" status
3. A researcher having correct access to the request action links relevant to them

# Out of scope

- The issue #347 also states to have the request list component display the correct requests given the context i.e. I'm a researcher, I will see my requests I created etc. The UI is prepared for this, but the backend does not return the correct data. This has been pushed to MauroDataMapper/sde-core#132 to fix

### DIFF
--- a/angular.json
+++ b/angular.json
@@ -77,7 +77,11 @@
               "optimization": false,
               "vendorChunk": true,
               "extractLicenses": false,
-              "sourceMap": true,
+              "sourceMap": {
+                "scripts": true,
+                "styles": true,
+                "vendor": true
+              },
               "namedChunks": true
             }
           },
--- a/src/app/secure-data-environment/components/sde-requests/sde-requests.component.html
+++ b/src/app/secure-data-environment/components/sde-requests/sde-requests.component.html
@@ -16,6 +16,4 @@ limitations under the License.
 
 SPDX-License-Identifier: Apache-2.0
 -->
-<p>sde-requests works!</p>
-
-<sde-requests-list></sde-requests-list>
+<sde-requests-list [useConfig]="requestsListConfig"></sde-requests-list>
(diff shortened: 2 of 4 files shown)


NigelPalmer (approved):
Small changes since all the heavy lifting is done in `sde-resources`. All looks good. Waiting on approval of associated tasks before merging.

PR #373, opened 2024-03-12
Feature/gh 366

### LINKED ISSUE #366: Query Builder: Limit Rule Sets to single table
# Description

The current Cohort and Data query builders provide too much flexibility which makes it hard to transform the generated MEQL into SQL. We want to restrict the Rule Sets that queries are built out of so that they are only applied to a single table. Also the root of the query builder should only allow ruleset to be added. 

# Additional context

The Query builder UI should be updated so that the items in red are removed and the items in green are added:
image

When the "Rule Set" button is clicked the user should be presented with an option as to which table they want to associate the Rule Set with. The selected table will then be used as the core table for that rule set. The user will only be able to select fields from that table for that ruleset. The users will be able to create a nested "Rule Set" for this "Rule Set". But this time when they click the nested "Rule Set" button they will only be presented with the tables that are referenced by the table in the parent "Rule Set". 

### Cohort Query
The Cohort Query will be tied to the core of the star schema. By default the user will be selecting fields related to the core table, but they will then be able to create "Rule Sets" that reference other tables.

### Data Query
At the base level, it is only possible to create "Rule Sets" in the Data Query. Each ruleset will be linked to a chosen data class. It is correct that the user can select the core table as a "Rule Set" within the data query.

# Technical
The query builder UI is generated using a third party component. We will need to work out how to apply these changes to it. This may not be straightforward.

NigelPalmer:
This work brings the query builder source code into our own code base so that we can manipulate it to fulfil the requirements of the MEQL to SQL process. To test his work you will need to reference the latest commits to the develop branch of  https://github.com/MauroDataMapper-Plugins/mdm-plugin-explorer . See https://github.com/MauroDataMapper-Plugins/mdm-plugin-explorer/pull/39 for how to build and include mdm-plugin-explorer in your mdm-core. You will need to set up some Profiles in Mauro to get the Query Builder to function correctly. But if you don't set up the Profiles first you should get some useful errors in the query builder telling you what needs to be done. How to setup a data model profile can be found here: https://github.com/MauroDataMapper-Plugins/mdm-plugin-explorer/pull/39 . The principal is the same for the primitive type profiles, you just need to make sure that you have a primitive type selected when creating a profile.

closes #366


src/app/data-explorer/query-builder-wrapper/query-builder-wrapper.component.html
```diff
@@ -26,50 +26,8 @@
       [config]="config"
       [operatorMap]="operatorMap"
       [allowRuleset]="allowRuleset"
+      [isDataQuery]="isDataQuery()"
```
abwilson23: When binding a method to a property in Angular, the method gets run upon every change detection cycle (possibly every keypress/mouse move/http event/etc) because Angular has to run the method in order to determine if it's value has changed. This is a relatively inexpensive method call, but it will still be run many more times than necessary. Instead best practice is to use a class property like `isDataQuery` because then Angular can do a simple 'has the value changed' check on it's own rather then run the method and then perform that check.

src/app/data-explorer/query-builder/dialogs/entity-selector-dialog/entity-selector-dialog.component.html
```diff
@@ -0,0 +1,50 @@
...
+</mat-dialog-content>
+<mat-dialog-actions align="end">
+  <button mat-stroked-button color="primary" (click)="close(false)">
+    {{ cancelLabel }}
+  </button>
+  <button mat-flat-button color="primary" (click)="close(true)">
+    {{ okLabel }}
+  </button>
+</mat-dialog-actions>
+-->
```
abwilson23: This necessary?

src/app/data-explorer/query-builder/pipes/arrow-format-pipe.ts
```diff
@@ -0,0 +1,29 @@
...
+SPDX-License-Identifier: Apache-2.0
+*/
+import { Pipe, PipeTransform } from '@angular/core';
+
+@Pipe({
+  name: 'arrowFormat',
+})
+export class ArrowFormatPipe implements PipeTransform {
+  transform(entityName?: string): string {
+    console.log(`arrowFormat: ${entityName}`);
```
abwilson23: Forgotten console.log

src/app/data-explorer/query-builder-wrapper.service.ts
```diff
@@ -65,18 +66,19 @@ export const mapOptionsArrayToModelDataType = (
 @Injectable({
   providedIn: 'root',
 })
-export class QueryBuilderService {
+export class QueryBuilderWrapperService {
   constructor(private profileService: ProfileService) {}
 
   public setupConfig(
```
abwilson23, line 81: This is quite a long and complicated method. Could you either break it into simpler private methods or add some more descriptive comments throughout please :).

src/app/data-explorer/query-builder-wrapper.service.ts
```diff
@@ -122,6 +194,36 @@ export class QueryBuilderService {
...
+    const requestOptions = {
+      handleGetErrors: false,
+    };
+
+    if (dataModel.id) {
+      return this.profileService.get(
+        CatalogueItemDomainType.DataModel,
+        dataModel.id,
+        'uk.ac.ox.softeng.maurodatamapper.plugins.explorer.querybuilder',
+        'QueryBuilderCoreTableProfileProviderService',
```
abwilson23: These strings are used in several places, could we move them into constants for easier maintanence.

abwilson23 (dismissed):
Some minor code comments left in files. Otherwise, changes look good overall.

**Functional testing**
1. If the type profiles are missing, there are many repeats of the error when one will suffice.
image
**_NP - Errors are no longer repeated_**
2. When copying a dataModel into the users data specifications, the query builder core table profile is not copied along with it which throws an error each time. 
**_NP - Fixed. This was quite a big change. The data model is created new so a profile needs to be copied to the new data model. Instead of doing this in the back end I have simply loaded the profile into the front end and then saved it to the new data model. This is the path of least resistance and since the profile is small I think this is an acceptable approach. Tests have been added to prove this works and some refactoring took place to get this implemented._**
3. DataQuery does not allow RuleSets from the core table. (Can't recall if this was in scope). It seems like it should allow data from the core table to be queried for. 
**_NP - Core table is now included in data query_**
4. Both Data and Cohort query allow Decimal inputs for Number type fields. (I'm not certain what the behaviour should be here, but this seems incorrect.)
image
**_NP - The input box in the screenshot above is not rendered correctly. This is because "Number" is not a valid type for the query builder. "Number" was replaced with "Integer" and "Decimal" over a year ago, see https://github.com/MauroDataMapper/mdm-explorer/issues/253 . I suspect that old primitive types have been used for this test and this is why the problem was seen. Correctly mapped Integer and Decimal types behave as expected._**
5. There is still an option to add a top level Rule in the cohort query despite the original ticket spec indicating that there should be no top level rules. The Data Query does follow the spec and only allows rulesets at the top level. 
image
**_NP - Cohort queries are allowed top level rules. This is correct behaviour._**
6. Changing between AND and OR in the rules/rulesets when editing a cohort/data query **does not** trigger the Save Changes box in the top right hand corner. In order for it to appear, one needs to actually edit one of the input fields. For example, starting with this on the Edit page
image
you can see that even after selecting AND instead of OR, the Save Changes does not display in the top right
image
only after editing the input box does the option to save appear. 
image
**_NP - Changing AND and OR now shows the save button_**

src/app/data-explorer/query-builder-wrapper/query-builder-wrapper.component.html
NigelPalmer: Changed to be a boolean variable

src/app/data-explorer/query-builder/dialogs/entity-selector-dialog/entity-selector-dialog.component.html
NigelPalmer: deleted

src/app/data-explorer/query-builder/pipes/arrow-format-pipe.ts
NigelPalmer: Deleted

src/app/data-explorer/query-builder-wrapper.service.ts
NigelPalmer: Method has been broken down.

src/app/data-explorer/query-builder-wrapper.service.ts
NigelPalmer: constants added

giovannaOCC:
Happy with the review. All seems to work as intended.
There are few more things raised and discussed, I'll open new issues for them.

giovannaOCC (approved):

PR #376, opened 2024-03-14
gh-375 - return the sde user details rather than the mdm user details.

### LINKED ISSUE #375: Return correct sdeUserDetails information
Ensure correct sdeUserDetails are returned in sdeSecurity.getSignedInUserDetails() method. 

- [ ] use newly created SdeUserDetails object. 
- [ ] retrieve correct information from local storage in implementation of previously mentioned method.

abwilson23:
Return the correct sde user information. Move object construction into userDetails service. 

closes #375


NigelPalmer (approved):
This rename make sense. **Passed**

PR #388, opened 2024-04-26
Feature/gh-382: rename submit to finalise with data specification

### LINKED ISSUE #382: Submit Data Specification - Rename Submit to Finalise
# Description

Data Specifications have a submit button:

image

Currently when clicked it will set the status of the Data Specification to be submitted which will change the Data Specification to be read only. We need all to change the meaning of this from Submit to Finalise. The submission will be done by an associated Data Request. 

1. Change the button label to read "Finalise and Finalise".
2. Change all function names, variables and any other bits of source code that may cause confusion to refer to finalise instead of submit.
3. Make sure that the badge that is displayed after the process has completed says "Finalised" rather than "Submitted"

giovannaOCC:
See issue #382 

NOTE: Renaming didn't affect css elements

### DIFF
--- a/src/app/data-explorer/data-explorer.types.ts
+++ b/src/app/data-explorer/data-explorer.types.ts
@@ -229,7 +229,7 @@ export const mapProfileSearchResult = (
   };
 };
 
-export type DataSpecificationStatus = 'unsent' | 'submitted';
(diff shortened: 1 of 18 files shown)


NigelPalmer (approved):
This is all working as expected and the changes look correct. *Passed*

PR #400, opened 2024-05-09
gh-387 - scaffold submission service

### LINKED ISSUE #387: Create Data Specification Submission Service
# Description

When the Submit button is clicked a number of steps need to be run. This new service will provided a scalable and customisable list of steps that will facilitate this process. The steps need to be written in such a way that if the process is run again then it will be able to continue from the last successful step. The steps need to be asynchronous and the solution needs to be coded to handle this.

Once this issue is completed the basic scaffolding for the service should be setup for the overarching task. That is, roughly the following files will be contained in a new DataSpecficationSubmission folder
1. A `SpecificationSubmissionService` created to ask as the main control service of the overall process. 
2. A `SpecificationSubmissionStateService` responsible for managing the state of the submission process. Should adhere to an interface which defines set, get, methods. Ideally store the state in an observable acting as a cache.
3. A `specification-submission.resource.ts` file created to contain the necessary types
4. An `ISpecificationSubmissionStep` interface defining the necessary logic for a step in the submission process. 
5. A test proving the scaffold works with a simple series of 1 or 2 steps. 

# Technical

### SubmissionStepService
Create an interface called ISubmissionStep that will be used as the base class for all submission step objects. This interface will look roughly like the following: 
``` ts
inteface ISubmissionStep {
  name: string;
  executionRequired(dataSpecificiationId: UUID): Observable<boolean>;
  execute(dataSpecificiationId: UUID): Observable<any>;
  saveNecessaryResults(newState: SubmissionState): void; // Set state
  validateExecutionInputs(): void;
}
```

~~### FinaliseStepService~~
~~The FinaliseStepService will be a class that implements SubmissionStepService. This will be the first class to be implemented to prove that these changes work correctly. More steps will be added in other issues. The FinaliseStepService will provide the following functionality:~~

- ~~name: "Finalise data specification"~~
- ~~executionRequired: Return true if the data specification that matches the passed in dataSpecificationId has NOT been finalised.~~
- ~~execute: Run the code that currently sits behind the "Finalise and Submit" button. This code will need to be moved into this class and the "Finalise and Submit" button action will need to instead call the SubmissionService.Submit function. This exercute function should return an empty key map. (I don't think we need to return any data from this step for downstream jobs to consume)~~
- ~~getResult: Return an empty key map.~~

### SubmissionService
The Submission service will have a hard coded list of objects that inherit from SubmissionSteps. Each item of the list will be iterated over in order and the results of each step will be passed into the next step. So this will look something like this (again, pseudo esq code)

So our list will look something like this for this task:
```
stepList = [new FinaliseStepService()]
```

The public submit function will initiate the recursive function and make the first call.
```
function submit(dataSpecificiationId: UUID) {
  // create an empty keymap
   submitStep(dataSpecificiationId, keyMap, steps, 0) // Some sort of observable call
}
```

Then the private recursive function will look something like this.
```
private function submitStep(dataSpecificiationId: UUID, keyMap: KeyMap, steps: Step[], index: number): Observable<StepResult> {
    if (index >= steps.length) {
        // Base case: if index is out of bounds, return an observable of a StepResult with no content
       return...
    }

      // Recursive case: execute the current step and then recursively call submitStep for the next step
     if (steps[index].executionRequired(dataSpecificiationId, keyMap) { // This if also needs to be called as an observable, so this line needs adapting
      return steps[index].execute().pipe(
          switchMap((stepResult: StepResult) => {
              // Update keyMap with the result of the current step
              const updatedKeyMap = stepResult.keyMap;
              // Recursively call submitStep for the next step
              return submitStep(steps, index + 1, updatedKeyMap);
          })
      );
     }
     else {
        return steps[index].getResult().pipe(
          switchMap((stepResult: StepResult) => {
              // Update keyMap with the result of the current step
              const updatedKeyMap = stepResult.keyMap;
              // Recursively call submitStep for the next step
              return submitStep(steps, index + 1, updatedKeyMap);
          })
     }
}
```

abwilson23:
- Create initial implementation of the overall submission process flow including: interfaces, main submission control flow, generic step stub and basic testing. 
- slight start to the select project step just to have a step to test. 

closes #387


NigelPalmer (changes requested):
This is looking like a good start, well done!

Now we have this in place, I would like to take this task a little further to prove that the submission service will behave in the way that we would like. To that end, please update the task with the following functionality:

1. Change the "Select Project Dialog" to be a dialog with a button on it that says "Click to complete step 1". When clicked the specification description should be updated with the text "Submission Step 1 Complete". 
2. Update `SelectProjectStep.run` so that the stepResult includes a string called `whatStep1Did`. Set `whatStep1Did` to the message "Step 1 ran this time"
3. Update `SelectProjectStep.isRequired` to check the specification description:
If the description includes the text "Submission Step 1 Complete",  set `isRequired` to false and set `whatStep1Did` to "Step 1 was not required to be run".
If the description _does not_ include the text "Submission Step 1 Complete", set `isRequired` to true 
4. Create a Step 2. Step 2 should simply show a dialog that displays the  `whatStep1Did` message.

Regarding code review, nothing major has leapt out yet, but I would rather carry out a more thorough code rerview once the above is complete.

Apologies for moving the goal posts, but I think this will be worth doing.

src/app/data-explorer/specification-submission/submission-steps/select-project.step.ts
```diff
@@ -0,0 +1,87 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+import { Injectable } from '@angular/core';
+import { Observable, map, of, switchMap, tap } from 'rxjs';
```
NigelPalmer: switchMap is not used. This will fail linting.

src/app/data-explorer/specification-submission/submission-steps/select-project.step.ts
```diff
@@ -0,0 +1,87 @@
...
+    const description: string = localStorage.getItem('description') ?? '';
+    const isRequired = !description.includes('Submission step 1 complete.');
+
+    const stepResult: StepResult = {
+      result: {
+        whatStep1Did: isRequired
+          ? 'SelectProjectStep is required.'
+          : 'Step 1 was not required to be run.',
+      },
+      isRequired: isRequired,
```
NigelPalmer: 'Expected property shorthand' error is being shown in my Visual Studio Code. Please change to just be isRequired.

src/app/data-explorer/specification-submission/submission-steps/select-project.step.ts
```diff
@@ -0,0 +1,87 @@
...
+      .afterClosed()
+      .pipe(
+        tap((result: SelectProjectDialogResponse) => {
+          localStorage.setItem('description', result.stepDescription);
+        }),
+        map(
+          () =>
+            ({
+              result: { whatStep1Did: 'Step 1 ran this time.' },
+              isRequired: false,
```
NigelPalmer: Why is isRequired being set to false here? This is confusing to follow. What is this isRequired even being used for after this point? Could we not just default isRequired to false in the StepResult and therefore not set the value here?

src/app/data-explorer/specification-submission/submission.resource.ts
```diff
@@ -0,0 +1,64 @@
...
+}
+
+export type StepName =
+  | 'SelectProject'
+  | 'GetDataRequest'
+  | 'GenerateSqlFile'
+  | 'AttachSqlFileToDataRequest'
+  | 'GeneratePdfOfDataRequest'
+  | 'AttachPdfToDataRequest'
+  | 'SaveDataRequest';
```
NigelPalmer: The idea of StepName was that it is something that would be displayed to the user when a submission is running. Therefore these should be sentence case rather than looking code like. e.g:
'Select project'
'Get Data Request'
'Generate SQL file'
etc

NigelPalmer (changes requested):
This was not quite what I had in mind, but the implementation clearly shows the step process working. Provided I type in "Submission step 1 complete." into the dialog, then the subsequent steps will fast track to step 2. If anything else is typed, then step 1 is run again. This is good and shows everything working as it should.

There are a couple of code observations that need addressing.

src/app/data-explorer/specification-submission/submission-steps/select-project.step.ts
abwilson23: Resolved.

src/app/data-explorer/specification-submission/submission-steps/select-project.step.ts
abwilson23: Resolved.

src/app/data-explorer/specification-submission/submission.resource.ts
abwilson23: Changed.

src/app/data-explorer/specification-submission/submission-steps/select-project.step.ts
abwilson23: I went with removing it altogether as only the result is used in the main submit method to save any new data to the state.

NigelPalmer (approved):
This framework is now ready to be populated with steps. This is all working well and a simple and flexible framework.

PR #405, opened 2024-05-24
Feature/gh-396 - create data request step

### LINKED ISSUE #396: Data Submission - Create Data Request
# Description

When the user clicks the "Submit" button, the first thing they should be shown is a dialog box asking them to select a Project to associate the Data Request with. They should only be able to view projects that they are allowed to see. The user should be able to cancel the process at this point. After a project has been selected by the user, then a new Data Request should be created.

For more info on the project selection portion of this task, see here: #394 

# Technical

### CreateDataRequestStepService
Create a new service called `CreateDataRequestStep` that inherits from `ISubmissionStep`. This should implement the following:

1. **isRequired**: Check to see if a Data request exists for the dataSpecificationId that is being submitted. (You will be able to call `/request-for-data-specification/{mauroDataSpecificationId}` to get this information, see maurodatamapper/sde-core#191 ) If there IS NOT a data request, then return true, otherwise, return false and the requestId for the dataSpecificationId that is being submitted. 
2. **run**: Extract the projectId from the SubmissionStateService. Get the Data Specification for the dataSpecificationId and get the Title and Content fields. Then create a new data request with the following information:
ProjectId = passed in project id
Title = Title from Data Specification
Content = Description from Data Specification. If there is no description then use the text 'no content'

abwilson23:
Implement the CreateDataRequest step of the submission service process.

- Add testing for the step
- Add basic custom error handling
- Adjust dialog box

closes #396


NigelPalmer (approved):
This is all looking good. The functionality is behaving as expected. The tests look good, and everything is working as it should be. Well done!

PR #406, opened 2024-06-11
Feature/gh 403 attach sql

### LINKED ISSUE #403: Data Submission - Attach SQL file
# Description

When the user clicks the "Submit" button, after the SQL representation of the request has been generated, that SQL file needs to be attached to the Data Request.

# Technical

### AttachSQLStep
Create a new step called `AttachSQLStep` that implement `ISubmissionStep`. This should implement the following:

1. **isRequired**: Get the requestId from the SubmissionStateService and retrieve that request. Check the attachments list on that request for an attachment. This check is also done in the previous step but we should do it again to be safe.
2. **run**: Get the requestId and pathToExportFile from the SubmissionStateService. Use these parameters to attach the file to the Data Request.

NigelPalmer:
Closes #403 

This PR has ended up including all the steps required to submit a data specification. The result of clicking the Submit button is:

1. A Data Request should exist with the title and content matching the title and description of the data specification
2. The Data Request should have a status of "Awaiting Approval"
3. The Data Request should have a SQL representation of the request attached to it
4. The Data Request should have a PDF representation of the request attached to it


OButlerOcc (dismissed):
So it works, and the code is readable (I had a parse through the tests and checked they work too) I found some bugs in my testing but am not 100% sure which pr they relate to so I've combiled them into a word doc and can go over them with you monday.

OButlerOcc (approved):
test fix

PR #409, opened 2024-06-21
Feature/gh 386 view request button

### LINKED ISSUE #386: Data Specification - View Request Button
# Description

After a data request has been created there should be a button that will link to the data request associated with the data specification. This should be on the right of the copy, share and version section. It would also be good to have some sort of divider. e.g: something like this.
image
When the "View Request" button is clicked the Request dialog should be displayed showing the associated Data Request.

Note that if a new version is created then the "View Request" button should only be displayed for the version that has been submitted. When the user changes the version number to display an unsubmitted request then the "View request" button should not be present.

# Additional Context
This work needs to build on maurodatamapper/sde-core#191 which will implement the functionality for linking the specification ids together. It will be required to update the functionality behind the submit button to make sure that the data specification id is set when the data request is created.

Once this work is completed there should be a data request that is linked to the data specification. This data request will not contain any information about the data specification. This will come later in another issue where we will auto attach that information as file attachments.

NigelPalmer:
closes #386 

This task has grown into improving the whole UX experience when submitting Data Specifications.

### DIFF
--- a/src/app/data-explorer/specification-submission/services/submission.broadcast.service.ts
+++ b/src/app/data-explorer/specification-submission/services/submission.broadcast.service.ts
@@ -1,0 +1,16 @@
+import { Injectable } from '@angular/core';
+import { BroadcastService } from 'src/app/core/broadcast.service';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class SubmissionBroadcastService {
+  constructor(private broadcastService: BroadcastService) {}
+
+  broadcast(message: string) {
+    this.broadcastService.loading({
+      isLoading: true,
+      caption: `Submitting your data specification - ${message}`,
+    });
+  }
+}
--- a/src/app/data-explorer/specification-submission/services/submission.broadcast.service.ts
+++ b/src/app/data-explorer/specification-submission/services/submission.broadcast.service.ts
@@ -1,3 +1,21 @@
+/*
+Copyright 2022-2023 University of Oxford
+and Health and Social Care Information Centre, also known as NHS Digital
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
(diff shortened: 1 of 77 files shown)


src/app/data-explorer/specification-submission/services/submission.broadcast.service.ts
```diff
@@ -0,0 +1,34 @@
...
+
+SPDX-License-Identifier: Apache-2.0
+*/
+import { Injectable } from '@angular/core';
+import { BroadcastService } from 'src/app/core/broadcast.service';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class SubmissionBroadcastService {
```
abwilson23: Just wondering why you didn't include this in the existing broadcast service?

src/app/pages/my-data-specification-detail/my-data-specification-detail.component.html
```diff
@@ -66,26 +69,36 @@ <h2>Queries</h2>
     <mdm-data-query-row
       queryType="cohort"
       [condition]="cohortQuery"
-      [readOnly]="dataSpecification.status === 'finalised' || isEmpty || !currentUserOwnsDataSpec"
+      [readOnly]="
+        dataSpecification.status === 'finalised' ||
+        dataSpecification.status === 'submitted' ||
+        isEmpty ||
+        !currentUserOwnsDataSpec
```
abwilson23: As this same boolean check is performed twice it might be nice to move this into a class variable.

src/app/data-explorer/specification-submission/services/submission.broadcast.service.ts
NigelPalmer: I thought the broadcast service was a more general service to be utilised. But I now see that there is stuff in their that is specific to users, so I have got rid of the SubmissionBroadcastService and moved the new code into the broadcastService.

abwilson23 (changes requested):
Functional testing has gone well, I can successfully create, submit, and view data requests from data specifications. There are two minor code comments which I will leave to you to resolve.

src/app/pages/my-data-specification-detail/my-data-specification-detail.component.html
NigelPalmer: I have created a pipe to handle this and added tests. It doesn't save a lot of code, but it does keep the logic in one place, and the tests could prove useful in the future.

abwilson23 (dismissed):
Review changes look good. Approved ✅

abwilson23 (approved):
An additional minor change was necessary. Approved ✅

PR #414, opened 2024-07-03
gh-413 - refactor createFromDataElements

### LINKED ISSUE #413: Minor refactor of createFromDataElements
As in title. 

Just need to verify that tests pass and creation of data specifications from data elements still functions correctly.

abwilson23:
Just make sure creating a data spec from data elements still functions correctly in the ui. 

closes #413

### DIFF
--- a/src/app/data-explorer/data-specification.service.spec.ts
+++ b/src/app/data-explorer/data-specification.service.spec.ts
@@ -484,12 +484,7 @@ describe('DataSpecificationService', () => {
         return cold('-a|', { a: undefined });
       });
 
-      submissionSDEServiceStub.mapToDataSpecification.mockImplementation((dataModel) => {
-        expect(dataModel.id).toBe(dataSpecification.id);
-        return dataSpecification;
-      });
-
-      const expected$ = cold('--------(a|)', {
+      const expected$ = cold('------(a|)', {
         a: dataSpecification,
       });
       const actual$ = service.createFromDataElements(elements, user, name, description);
--- a/src/app/data-explorer/data-specification.service.spec.ts
+++ b/src/app/data-explorer/data-specification.service.spec.ts
@@ -484,7 +484,12 @@ describe('DataSpecificationService', () => {
         return cold('-a|', { a: undefined });
       });
 
-      const expected$ = cold('------(a|)', {
+      submissionSDEServiceStub.mapToDataSpecification.mockImplementationOnce((dataModel) => {
+        expect(dataModel.id).toBe(dataSpecification.id);
+        return dataSpecification;
+      });
+
+      const expected$ = cold('-------a|', {
         a: dataSpecification,
       });
       const actual$ = service.createFromDataElements(elements, user, name, description);
--- a/src/app/data-explorer/data-specification.service.ts
+++ b/src/app/data-explorer/data-specification.service.ts
@@ -546,8 +546,12 @@ export class DataSpecificationService {
       this.create(user, name, description),
     ]).pipe(
       switchMap(([rootDataModel, dataSpecification]) => {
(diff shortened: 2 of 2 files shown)


src/app/data-explorer/data-specification.service.ts
```diff
@@ -542,34 +541,44 @@ export class DataSpecificationService {
...
       // TODO: assume there is only one data model, will have to change in future
+      // Get the rootModel and create a new empty data specification.
       this.dataExplorer.getRootDataModel(),
       this.create(user, name, description),
     ]).pipe(
       switchMap(([rootDataModel, dataSpecification]) => {
-        if (!rootDataModel || !rootDataModel.id) {
-          return throwError(() => new Error('No root data model'));
+        if (!rootDataModel?.id || !dataSpecification?.id) {
+          return throwError(() => new Error('Missing root data model or data specification'));
```
NigelPalmer: I don't like this. This makes it harder to work out what has gone wrong. Is it broken because the root model is missing or because the data specification is missing? I would prefer this check as separate errors: e.g:
```
        if (!rootDataModel?.id) {
          return throwError(() => new Error('Missing root data model'));

        if (!dataSpecification?.id) {
          return throwError(() => new Error('Missing data specification'));
```

src/app/data-explorer/data-specification.service.spec.ts
```diff
@@ -484,12 +484,7 @@ describe('DataSpecificationService', () => {
         return cold('-a|', { a: undefined });
       });
 
-      submissionSDEServiceStub.mapToDataSpecification.mockImplementation((dataModel) => {
-        expect(dataModel.id).toBe(dataSpecification.id);
-        return dataSpecification;
-      });
```
NigelPalmer, line 490: Not clear as to why this has been removed from the test?

src/app/data-explorer/data-specification.service.ts
```diff
@@ -542,34 +541,44 @@ export class DataSpecificationService {
...
+      map(({ populatedDataSpecification, coreTableProfile }) => {
+        // Save the core table profile to the newly created data specification.
+        this.coreTableProfileService.saveQueryBuilderCoreTableProfile(
+          populatedDataSpecification,
+          coreTableProfile
+        );
+
+        // We don't care about the output above so just return the data specification.
+        return this.submissionSDEService.mapToDataSpecification(populatedDataSpecification);
+      })
```
NigelPalmer, line 584: This is not working? When I try and create a specification and go to look at the query builder I get the error:
```
"Mauro Data Explorer - Query Builder Core Table" (data model profile) for DataModel validation errors:
- This field cannot be empty
- Format needs to be [Schema].[Table] e.g: people.patient
```
I checked the network log and I can see the profile being "got", but I cannot see it being saved.

NigelPalmer (changes requested):
A few coding issues and I couldn't get it to work?

src/app/data-explorer/data-specification.service.ts
abwilson23: Reverted.

src/app/data-explorer/data-specification.service.ts
abwilson23: This is due to the fact that I was incorrectly using `map` here. The fix is to use `switchMap` to switch to the observable emitted by the `saveQueryBuilderCoreTableProfile()` method and then pipe off the end of that to map the dataSpecification properly. This ensures the save call completes successfully before moving along to the next step.

src/app/data-explorer/data-specification.service.spec.ts
abwilson23: Neither..! Added back.

NigelPalmer (approved):
This now all seems to be working as it should.

PR #415, opened 2024-07-04
gh-412 - Add user friendly dialog if submission of data specification fails. 

### LINKED ISSUE #412: User friendly error when Data Submission fails
# Description

When a Data Submission is made, it may error for some reason, i.e it can't connect to azure to upload a file. When this happens a horrible error message is displayed to the user. What we need to do it catch all errors and give the user a nice friendly error message. It would also be useful for the real error message to be written somewhere, maybe the console log for now, so that it can be viewed when things go wrong.

The friendly error message should say "Submission Step _step name_ failed. Please press the Submit button again to submit your data specification. If you keep seeing this message then please contact Mauro administrators for help" With an OK button on it.

UPDATE: 
- Add contextual error messages to the submission service
- Update NoProjectsFoundError to match screenshot from nigel. 
- Also fixed problems with submission status dialog where it would only appear on the first submit click and the UI was see through so it was hard to see when there was not a white background.

# Additional context

The simplest way to break a submission is to turn off your local Azurite server if you are using that. Your other option is to corrupt the Azure references in the run xml files so that the Azure connection will fail.

abwilson23:
- Add `SimpleDialogComponent` for generic dialog messages. 
- Add a test checking the dialog is opened if an error is thrown during a test. 

Simplest way to test is to manually throw an error in one of the existing steps. I suggest a later step just to see how the UX flows from working to failure. 

closes #412

### DIFF
--- a/src/app/data-explorer/specification-submission/services/specification-submission.service.ts
+++ b/src/app/data-explorer/specification-submission/services/specification-submission.service.ts
@@ -32,6 +32,10 @@ import { AttachPdfStep } from '../submission-steps/attach-pdf.step';
 import { SubmitRequestStep } from '../submission-steps/submit-request.step';
 import { BroadcastService } from 'src/app/core/broadcast.service';
 import { DialogService } from '../../dialog.service';
+import {
+  DEFAULT_ERROR_MESSAGE,
+  NoProjectsFoundError,
+} from '../type-declarations/submission.custom-errors';
 
 @Injectable({
   providedIn: 'root',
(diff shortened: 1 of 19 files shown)


src/app/data-explorer/specification-submission/services/specification-submission.service.ts
```diff
@@ -116,12 +115,17 @@ export class SpecificationSubmissionService {
...
+    console.error(`Error running step ${stepName}. Step failed with error message: ${error}`);
+
+    const errorMessage = this.buildErrorMessage(stepName);
+    this.dialogService.openSimple({ heading: 'Submission Error', message: errorMessage });
+  }
+
+  private buildErrorMessage(stepName: StepName): string {
+    return `<p>Submission Step <b>${stepName}</b> failed.</p>
+    <p>Please press the Submit button again to submit your data specification.
+    If you keep seeing this message then please contact Mauro administrators for help.</p>`;
```
NigelPalmer: We are no longer showing a useful error message to the user. I just tried to submit and the user facing message told me to simply try again. Where as the console message said "NoProjectsFoundError: You do not have any projects to submit this data request for." What we need to do is display the errors that we know are user friendly, otherwise default to the current behaviour of "Please try again". I.e `if NoProjectsFoundError then display error message in dialog`

It should look something like this:
image

You may need to update the returned error message to include the additional text. Also there may be other exception types that we should be displaying the error message for.

NigelPalmer (changes requested):

src/app/data-explorer/specification-submission/services/specification-submission.service.ts
abwilson23: NoProjectsFoundError now produces a friendly message explaining how the user can action the issue on their own. All other errors default to a standard error message from the original ticket as the fixes are outside the control of the user. 
Simple to add further modifications to error handling if necessary.

NigelPalmer (approved):
User receives a clear message telling them what they need to do. We will likely need to update the wording in the future as the workflow changes, but the message we have now is accurate for the current state of the system.

PR #416, opened 2024-07-05
gh-399 - ensure new version only created upon clicking 'yes'.

### LINKED ISSUE #399: Data Specification: New Version always created after button click
# Description

After a Data Specification has been finalised, then the user can create a new version of the Data Specification. A confirmation dialog is shown. Regardless as to whether the user clicks "Yes" or "No" on that dialog, the new version is still created.

# Steps to reproduce

1. Go to 'My Data Specifications'
2. Click on a data specification
3. Click on Finalise and click 'Yes' on the prompt
4. Click on 'Continue' on the next prompty
5. Click on 'New Version'
6. Click on 'No' on the prompt, the new version is created anyway.

# Expected behavior

When the user is presented with "Do you want to create a new version of this data Specification? You will be redirected to the new version details page" and they click "No" then they stay on the page they are on and a new version IS NOT created.

# Screenshots

image

abwilson23:
- Make use of the dialogResponse.result instead of just checking the response object is initialized.

closes #399


NigelPalmer (approved):
Perfect! Simple change, achieves exactly what we need,

PR #418, opened 2024-07-10
gh-372 - auto add core table elements during copy

### LINKED ISSUE #372: Data Specifications: Empty core data class created
# Description

When creating a data specification the researcher can choose which data classes they want to include from the data model. The data model is expected to be a star schema, with the core data class identified by a data model profile. If the researcher only chooses all the data elements of a non core data class, then the data specification will also include the core data class. This is presumably because it is being referenced by a foreign key in from the chosen data class, and the mdm-core has worked out that this data class is required. What the data specification DOES NOT include is any of data elements in that core data class. At the very least it should include the fields that are being referenced by the foreign key. 

If a researcher selects a single data element from a non core data class then only the selected data class will be added to the data specification, and not the core data class. This is also wrong, since the core data class is always required for writing queries.

# Steps to reproduce

1. Go to the MDE
2. Click Browse
3. Select a schema
4. Select a data class
5. Click the 3 dots next to the data class and select "Copy to new specification"
6. Then open the specification
7. Scroll down to select the fields. The core data class will be empty

# Expected behaviour

I would expect the data specification to always include the core data class. At the very least the core data class should include the primary key data element. When selecting a non core data class the foreign key data element back to the core data class should always be included.

This is important, because without a core data class the query builder functionality will not work.

# Technical

The MDE needs to be updated to identify and add the additional tables and elements that are required before calling the mdm-core to copy the data model to the data specification. How we identify this data needs to be worked out but these rules should be applied:

1. Every Data Specification must include the core data class and the primary key data element of that data class
2. Every data class in a Data Specification that is not the core data class must include a foreign key data element that references the core data class. 

Depends on: maurodatamapper-plugins/mdm-plugin-explorer#49

**Complete when:** 
1. Any copying operation done between data specifications (including creation of a new data specification) done in mdm-explorer results in the aforementioned data elements being added to the target.

abwilson23:
Ensure that any time a data specification is copied or created that the primary key data element of the core table and any foreign keys to the core table from selected dataclasses are also copied over. This ensures the cohort/data queries function properly. 

- Add plumbing to connect to newly added mdm-plugin-explorer endpoint
- Update tests to ensure duplicates aren't copied/added 
- Prefer dataModel.id over full dataModel object reference for method parameter. 

Needs to be tested alongside: maurodatamapper-plugins/mdm-plugin-explorer#49

closes #372

### DIFF
--- a/src/app/data-explorer/core-table-profile.service.spec.ts
+++ b/src/app/data-explorer/core-table-profile.service.spec.ts
@@ -101,7 +101,7 @@ describe('CoreTableProfileService', () => {
         a: expectedResult,
       });
 
-      const actual$ = service.getQueryBuilderCoreTableProfile(rootDataModel as DataModel);
+      const actual$ = service.getQueryBuilderCoreTableProfile(rootDataModel.id);
 
       expect(actual$).toBeObservable(expected$);
     });
@@ -159,7 +159,7 @@ describe('CoreTableProfileService', () => {
         a: expectedFrameResults,
       });
 
-      const actual$ = service.getQueryBuilderCoreTableProfile(rootDataModel as DataModel);
+      const actual$ = service.getQueryBuilderCoreTableProfile(rootDataModel.id);
 
       expect(actual$).toBeObservable(expected$);
     });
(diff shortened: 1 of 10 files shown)


NigelPalmer (approved):
This is all working as expected. Well done!

PR #434, opened 2024-09-03
GitHub Actions build script

jamesrwelch:
Offering this for review.  The 'publish artifacts' step is not tested yet, but will be when we merge to develop - it uses the same code as mdm-ui which works.


NigelPalmer (approved):
The script looks correct and is running successfully. It will be reliant on merging mdm-resources and sde-resources to develop before the script will pass for some work, but that was always the way things were in Jenkins and there is no easy way to manage this. Overall, all looking great!

PR #443, opened 2024-09-10
gh-417 Clear selection when adding to a data spec or after 24h

### LINKED ISSUE #417: Elements already selected upon initial page load of dataclass details
$${\color{red}Please \space read \space comments \space before \space commencing \space task}$$

# Description

After viewing the details of a dataclass from the Browse page I will occasionally notice that the 'Current selection' section of the page indicates that an element(s) has already been selected. If I look through all the elements listed on the page, I can find no such selected element(s). 

# Steps to reproduce

1. Go to 'Browse' page
2. Select a Schema and then a DataClass and click 'View Details' in the bottom right. 
3. Inspect the 'Current selection' box in the left column of the page and notice that it indicates an element(s) is selected even though you haven't interacted with any of the data elements on the page yet. 

# Expected behavior

The number of selected elements to be 0 upon initial page load. 

# Screenshots

image

greg-at-occ:
As discussed with Nigel, doesn't quite match the ticket but this approach should strike a good balance of convenience and avoiding confusion.

Closes #417

### DIFF
--- a/src/app/data-explorer/selection-compact/selection-compact.component.html
+++ b/src/app/data-explorer/selection-compact/selection-compact.component.html
@@ -16,10 +16,7 @@ limitations under the License.
 
 SPDX-License-Identifier: Apache-2.0
 -->
-<div
-  class="highlight-box mdm-selection-compact"
-  *ngIf="(selectedElements$ | async)?.length != 0"
->
+<div class="highlight-box mdm-selection-compact" *ngIf="(selectedElements$ | async)?.length != 0">
   <div class="mdm-selection-compact__header">
     <h3>Current selection</h3>
   </div>
@@ -42,6 +39,7 @@ SPDX-License-Identifier: Apache-2.0
       <mdm-data-element-multi-select
         [sourceTargetIntersections]="sourceTargetIntersections"
         [dataElements]="(selectedElements$ | async) ?? []"
+        (selectionAddedToSpec)="clearSelection()"
       ></mdm-data-element-multi-select>
     </div>
   </div>
--- a/src/app/data-explorer/selection.service.ts
+++ b/src/app/data-explorer/selection.service.ts
@@ -21,6 +21,11 @@ import { Uuid } from '@maurodatamapper/mdm-resources';
 import { BehaviorSubject, Observable, Subject } from 'rxjs';
 import { DataElementSearchResult } from './data-explorer.types';
 
+type StorageType = {
+  lastUpdated: number;
+  selection: DataElementSearchResult[];
+};
+
 @Injectable({
   providedIn: 'root',
 })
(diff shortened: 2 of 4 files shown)


src/app/data-explorer/selection-compact/selection-compact.component.html
```diff
@@ -42,6 +39,7 @@ <h3>Current selection</h3>
       <mdm-data-element-multi-select
         [sourceTargetIntersections]="sourceTargetIntersections"
         [dataElements]="(selectedElements$ | async) ?? []"
+        (selectionAddedToSpec)="clearSelection()"
```
NigelPalmer, line 42: `mdm-data-element-multi-select` is used in other parts of the code and the `(selectionAddedToSpec)` has not been added to these definitions which means the selection IS NOT cleared when using those parts. E.g, this one does not clear the spec list after being clicked.

image

**Extra Task:** The dialog in the screenshot above also does not disappear after the spec has been updated, even if the user clicks on "Open Data Specification" after the fields have been added. Could we please fix this whilst we are working in this area.

src/app/data-explorer/selection.service.ts
```diff
@@ -35,14 +40,22 @@ export class SelectionService {
...
   private STORAGE_KEY = 'elementSelection';
+  private SELECTION_TIMEOUT = 24 * 60 * 60 * 1000;
 
   constructor() {
+    let storedSelection: DataElementSearchResult[] | null = null;
     const storedJson = localStorage.getItem(this.STORAGE_KEY);
-    if (storedJson == null) {
+    if (storedJson != null) {
+      const stored = JSON.parse(storedJson) as StorageType;
+      if (stored.lastUpdated + this.SELECTION_TIMEOUT > Date.now()) {
```
NigelPalmer, line 50: **Passed:** Selected items are being removed after 24 hours.
I made a small code change in my local copy to test this. The change I made was this:
```
      const testTime = Date.now() + this.SELECTION_TIMEOUT - 10000;
      if (stored.lastUpdated + this.SELECTION_TIMEOUT > testTime) {
```

I then ticked items and refreshed the page repeatedly. After 10 seconds the checks were removed. Therefore this is working as it should be.

NigelPalmer (changes requested):
### Issues
Selected items are not always removed after a Data Specification is updated or created. See comments for details.

### Passed
Time out of selected items is working correctly

src/app/data-explorer/selection-compact/selection-compact.component.html
greg-at-occ: Added the clearSelection trigger here, but not on bookmarks page, since that selection is not persistent anyway and user may find it helpful to add a small set of fields to multiple specs (not now possible from the main browse)

Extra Task seems to have resolved itself by virtual of clearing the selection

NigelPalmer (changes requested):
Unfortunately I have found another problem. If I log in as researcher_one and select some items, logout and login as manager_one then those items are still selected. I know it is a lot less common for people to share a computer these days, but this still does not feel correct. It creates the impression that users can view other users data.

greg-at-occ:
As discussed, use of localStorage can remain as-is for now, pending a larger review of its use in these UIs

NigelPalmer (approved):
Happy with this now. A wider review will be carried out later.

PR #466, opened 2024-10-10
gh-462: Improved Query Builder UX

### LINKED ISSUE #462: Query Builder - List picker control is messy
# Description

The query builder list picker control is messy when the options are verbose and contain lots of text. They also truncate too much information. See the screenshots below .

You need to identify the component that is being used and then update it so that the data is presented more cleanly. There is a lot of white space to the right of the component, we could see if we could make use of this.

# Screenshots
image

image

NigelPalmer:
closes #462


giovannaOCC (approved):
It's working as intended

PR #472, opened 2024-10-14
gh-470: Allow for duplicate field names

### LINKED ISSUE #470: Query Builder: Duplicate field names get removed from query builder
# Description

If there are data elements that have the same name that do not appear 

# Steps to reproduce

1. In Mauro Data Mapper create two data elements with the same name on two different classes.
2. Create a new data specification that includes both those data classes.
3. On that data spec create a cohort query. Try and select the field with the same name. Only one of them will be included.

# Expected behaviour

Both fields should be available to be selected.

# Technical
https://github.com/MauroDataMapper-Plugins/mdm-plugin-explorer/issues/59 needs to be tested and released alongside this change.

Existing cohort and data queries will cause errors in the system after this change has been deployed. This can be fixed by using the MDM UI and editing the cohort and data query rule representations. e.g:
```
      "entity": "Inpatient.INP_GENERAL",
      "field": "PERSON_ID",
```
needs to be updated to be:
```
      "entity": "Inpatient.INP_GENERAL",
      "field": "Inpatient.INP_GENERAL.PERSON_ID",
```

### LINKED ISSUE #471: EnumerationValue display in query builder page should show values rather than keys
# Description

Using an EnumerationType column in a query criteria, select `in` or `not in` as the criteria type. The `key`s of the EnumerationValues are shown rather than the `value`s. However the `value`s are shown when `=` is selected.

# Expected behavior

`value`s should always be displayed as the human readable option.

# Environment

latest of `feature/demo`

NigelPalmer:
closes #470 
closes #471

### DIFF
--- a/src/app/data-explorer/pipes/meql.pipe.spec.ts
+++ b/src/app/data-explorer/pipes/meql.pipe.spec.ts
@@ -43,7 +43,7 @@ describe('MeqlPipe', () => {
         condition: 'and',
         rules: [
           {
-            field: 'field name',
+            field: 'some entity.field name',
             entity: 'some entity',
             operator: '=',
             value: 'String 1',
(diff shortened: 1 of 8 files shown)


joe-crawford:
Have tested this branch and can see both issues are resolved, thanks @NigelPalmer!

Will wait to merge as there are failing tests. But if time is an issue happy to merge to the demo branch for now.

NigelPalmer:
Just working on the tests now. Hopefully quick turn around, then I can merge it in.

joe-crawford (approved):
Looks great :+1:

PR #485, opened 2024-12-04
Feature/gh 439 choose data spec

### LINKED ISSUE #439: Choose Data Specification for a request
# Description

#438 sets the Data Specification to a hard coded value. This needs to be enhanced so that the user is presented with a list of finalised Data Specifications and they get to choose the one that they want to submit. A dialog should be shown with a list and when selected the Data Specification description should be displayed. Once selected the ID and name of the Data Specification should be broadcast to the SDE.

The end result is that the user should now be able to choose any finalised Data Specification they wish to link to a New Project Request or Project Update Request, and the name of that Data Specification is shown on the Request.

NigelPalmer:
closes #439

Added a new step process to allow Data Specifications to be attached to requests. This is then injected into sde-resources so that it can be utilised within Request dialogs. Also updated the submission button in the Data Specification page so that a Wizard is now displayed that allows the user to pick and choose what type of request they want to attach the Data Specification to.


giovannaOCC:
Issue:

As a researcher, I'll go to "Browse the Catalogue" and 
- select some data (eg: Results -> Labs Details)
- then pick "Critical High" and "Critical Low"
(I guess any data can be selected, as long as they are "a lot").

Then, with the selected data, create a new Data Specification...
you will be stuck here:

image

The problem is that you are stuck with this BUT the data specification is indeed CREATED!
By refreshing the browser you can see, it is correctly created.

giovannaOCC:
Issue:

From "My Data Specification" I want to add a Finalised data to a "Project Change Request".
**If the request already exist:**
- I can do all steps for adding
- the UI respond consistently
- at the final confirmation, it seems to save, BUT it actually didn't.
=> my data specification is still "Finalised" and not "Assigned to a request"
=> the request I was using, doesn't have the data saved (attachments  is empty and the audit table in DB record the action but with MauroDataSpecification = null) -> it's indeed not showing the message that it is creating the PDF at all

**If the request is created at the moment:**
- I can follow all the steps until the Request Dialog 
- I can fill all the data required
- When I'm going to Save, the user portal crash with API communication problem and this further message:
image

Note: this steps have been done with 'just created' data specifications.

giovannaOCC (changes requested):
I've found a couple of issues that I've added in the Conversation area

NigelPalmer:
I have fixed the loading dialog issue.

I'm not convinced the project change one is a real issue. The error in the screenshot relates to mdm-core and this is to do with something going wrong with the dataSpecification itself. I think this is outside the scope of attaching.

giovannaOCC:
> I have fixed the loading dialog issue.
> 
> I'm not convinced the project change one is a real issue. The error in the screenshot relates to mdm-core and this is to do with something going wrong with the dataSpecification itself. I think this is outside the scope of attaching.

it's indeed something belonging to mdm-core.
Further message I could get from mdm-core console while I was re-attempt the steps described above:

`Caused by: org.hibernate.StaleStateException: Batch update returned unexpected row count from update [0]; actual row count: 0; expected: 1; statement executed: update openidconnect.discovery_document set version=?, date_created=?, last_updated=?, path=?, token_endpoint=?, end_session_endpoint=?, userinfo_endpoint=?, created_by=?, authorization_endpoint=?, jwks_uri=?, issuer=? where id=? and version=?`

and

`Caused by: uk.ac.ox.softeng.maurodatamapper.api.exception.ApiInternalException: Cannot store an unlocked GroupBasedUserSecurityPolicyManager`

giovannaOCC (approved):
all works as expected

PR #489, opened 2024-12-10
Visual tweaks

### LINKED ISSUE #454: Data Specifications: only 10 columns per table are displayed when viewing
# Description

When viewing a data specification at a URL like `http://localhost:4201/#/dataSpecifications/{{uuid}}`, only 10 data elements per selected data class are displayed when viewing, even if more were selected.
# Steps to reproduce

Using a root data model which contains a table data class with more than 10 column data elements, build a query selecting all the elements on that data class. When viewing the data specification, only 10 data elements show under that data class.
# Expected behavior

There should be a way to view more/all of the data elements selected.
# Screenshots

`inpatient_encounter` had more than the displayed 10 data elements selected:

Screenshot 2024-09-24 at 4 31 05 PM

# Environment

latest develop of `mdm-explorer` and `mdm-plugin-explorer`

### LINKED ISSUE #495: HTML descriptions are not handled
# Description

In the Mauro Data Mapper, descriptions can be written using HTML. These descriptions are displayed as raw HTML in the User Portal.

# Steps to reproduce

1. Open Mauro Data Mapper
2. Edit a description in HTML mode and add some colours and font changes
3. View the description in the User Portal. It will be displayed as raw HTML.

# Expected behaviour

The descriptions should appear like they do in the Mauro Data Mapper. See screenshots below for comparisons.

# Screenshots

### MDM Description
image

### User Portal Description
image

joe-crawford:
Some visual and other tweaks from the demo in October:

* Allow adding HTML in descriptions, useful for the documentation we're importing from source systems
* Tweak header and login button padding to make the same as sde-admin-ui (have another PR to raise there to make it the same)
* On data specification page, hard code max limit on dataElements request (fixes #454) and sort elements, avoids a backend change for now


joe-crawford:
Resolves https://github.com/MauroDataMapper/mdm-explorer/issues/495

NigelPalmer (approved):
This all looks good and has improved the UX. Approved.

NigelPalmer (approved):
This is all looking good. There is still an issue with HTML where colours are not being applied but I don't think it is worth holding up this task for that. We can address this if it really becomes an issue in the future. Approving.

PR #490, opened 2024-12-10
Allow OpenID Connect login from Micronaut

### LINKED ISSUE #494: Allow OpenID Connect login from Micronaut
See associated PR for full details

joe-crawford:
This PR makes the same change as in the main UI in https://github.com/MauroDataMapper/mdm-ui/pull/912 in order to allow login using OpenID Connect when using mauro-micronaut instead of mdm-core, which needed some tweaks as we're using Micronaut's provided OpenID plugin rather than our custom implementation.

Raising this PR even though we're still working on mauro-micronaut as want to keep all the UI code in one codebase.


joe-crawford:
sorry @pjmonks requested your review by accident!

src/app/security/open-id-connect-authorize/open-id-connect-authorize.component.ts
```diff
@@ -87,45 +88,73 @@ export class OpenIdConnectAuthorizeComponent implements OnInit {
...
+          providerId,
+          state,
+          sessionState,
+          code,
+        })
+        .pipe(
+          catchError((error: LoginError) => {
+            switch (error.type) {
+              case SignInErrorType.InvalidCredentials:
+                this.errorMessage = 'Invalid username or password!';
```
NigelPalmer, line 125: I'm not sure how this would occur? If you have invalid credentials you cannot log into the OpenID connect provider, so you are not logged in and unable to get here. Unless there is another way to get to this code?

src/app/security/open-id-connect-authorize/open-id-connect-authorize.component.ts
```diff
@@ -87,45 +88,73 @@ export class OpenIdConnectAuthorizeComponent implements OnInit {
...
+                break;
+              case SignInErrorType.AlreadySignedIn:
+                this.errorMessage = 'A user is already signed in, please sign out first.';
```
NigelPalmer, line 128: Likewise if a user is signed in then you will automatically sign in as that user when you go to the User Portal, so is there ever a circumstance where you can be logged in as someone else?

NigelPalmer (changes requested):
I was able to sign in successfully and everything works as it should. I have raised some questions about the errors, but I don't think these questions are a reason not to merge this code.

I have not tested this against a micronaut implementation. I am just confirming that this all works fine with an mdm-core backend.

Reopening so questions can be answered.

src/app/security/open-id-connect-authorize/open-id-connect-authorize.component.ts
joe-crawford: Yes I agree with both of the comments, I think you'd have to get into an inconsistent state between Explorer and the Identity Provider for any of these to happen, however I haven't added these messages, just moved them around in the code... I'm guessing if they ever did appear they might be useful for debugging so happy to leave these in for now?

NigelPalmer: Yes, lets leave them in. They are benign and I don't think it is worth the time and effort to remove them. And as you say, it may help with debugging.

NigelPalmer (approved):
Happy with the responses to my questions. This is now passed.

Issue #110, opened 2022-05-31
Two users cannot create a data specification with the same name


pjmonks:
# Description

Due to all data models needing a unique label in the metadata catalogue, it is not possible to create a data specification that has the same name as another model/specification in the same catalogue. This is good for one user preventing them from duplicating their specification names, but it hinders different users from unknowingly creating specifications with the same name.

# Steps to reproduce

1. Sign in as one user to MDE.
2. Go to the Browse page
3. Hover over one schema or data class, click on the "..." then click "Copy to new specification"
4. Follow the instructions in the dialog, using the specification name "test specification"
5. Ensure the specification was created successfully
6. Sign out and sign in again as a different user
7. Repeat steps 2 to 4 - use the _exact same_ name for the specification.

# Expected Behaviour

Multiple users should not know about other user's specification, so they should be able to create specifications with no limitations on name.

# Additional Context

This is the error that comes back from Mauro:

```
# Request (sample)
POST api/folders/{folderId}/dataModels

{"label":"My first specification","description":"","type":"Data Asset","folder":"375bc6ae-e776-46ec-abd0-4c3fa921ca22","author":"Research Test","organisation":"OCC"}

# Response
{
    "total": 1,
    "errors": [
        {
            "message": "Property [label] with value [My first specification] must be unique by branch name"
        }
    ]
}
```

Preventing duplicate labels in the catalogue is by design, but this becomes a user experience issue for researchers. @jamesrwelch , @olliefreeman - is there some alternative we can come up with?

pjmonks:
This issue needs more discussion with @jamesrwelch for the best approach. This may also be related to other areas of organising data requests (e.g. "projects", "organisations" etc), so it might be best to discuss this then.

pjmonks:
I would suggest this issue be put on hold until the SDE Administration is better defined. Changing how the data requests work may affect both the explorer and the admin interfaces.

PR #491, opened 2024-12-10
gh-441: Provide more useful error message when there is a data specif…

### LINKED ISSUE #441: Create Data Specification - Errors not being reported to the user
# Description

If I create a data Specification with the same name as another Data Specification I get the error
"Property [label] of class [class uk.ac.ox.softeng.maurodatamapper.datamodel.DataModel] with value [Test Copy] must be unique by branch name" from the back end, but the front end does not report this problem. We need to present the user with a more helpful error message so they know what they have done incorrectly.

NigelPalmer:
…ication naming clash

closes #441 

Note that this issue is to create a solution to simply present the error in a more helpful way. #110 has been created to deal with this problem in a way that is more beneficial for the users.

### DIFF
--- a/src/app/data-explorer/data-specification.service.ts
+++ b/src/app/data-explorer/data-specification.service.ts
@@ -625,9 +625,13 @@ export class DataSpecificationService {
             of(dataElements),
           ]);
         }),
-        catchError((error) => {
+        catchError((error: HttpErrorResponse) => {
+          const errorMessage = this.isDataSpecificationNameClashError(error)
+            ? 'A data specification with this name already exists. Please choose a different name.'
+            : error.message;
+
           this.toastr.error(
-            `There was a problem creating your data specification. ${error}`,
+            `There was a problem creating your data specification. ${errorMessage}`,
             'Data specification creation error'
           );
           return EMPTY;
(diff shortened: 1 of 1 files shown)


giovannaOCC (commented):
Perhaps I'm not getting the problem:

- I've created a Data Specification named "TestRenaming"
- I finalised "TestRenaming"
- From finalised "TestRenaming", I create a copy and I name it "TestRenaming"

what I got is a Data Specification, named "TestRenaming" in Draft mode.

Shouldn't I get the error message?


--- EDIT
It does works if I create a the data specification directly from catalogue, but not from "copy".
Furthermore, the copy with the same name replace di existing one, which I believe it is wrong.

NigelPalmer:
After further testing what is happening is this:
1. If the name clashes with an existing Draft data specification then the error is displayed.
2. If the name clashes with an existing Finalised data specification then a new data specification is created with the same name. This is actually a new version of the data specification with the same name and when finalised a new version is created. But the User Portal then gets confused and shows this as version 1 and not the correct version number. If then the "Create new version" button is clicked then a new version is created with the correct version number, but it is only possible to see that one and the previous version, and not all the versions.

The correct way to solve this is to prevent the duplicate name from being passed to the create routine. This is what #110 is to address, so I suggest we pass this task based on the message works when the clash occurs, and then I will schedule #110 to get this fixed properly.

giovannaOCC (approved):
I can confirm all the behaviour you have described. I'm happy with this fix for improving the error message in the scenario described as well as the followup task for handling the duplicated name separate issue.

PR #493, opened 2024-12-11
gh-481: Turn off Data Request auto submission

### LINKED ISSUE #481: Turn off auto submission for data requests
# Description

When a data specification is submitted then the data request is created and automatically submitted. We need to change this so that it is in line with the other request types, i.e only the request is created and it is up to the user to save and submit it.

# Technical

Change the Data Request submission process to use the same steps as the other requests.

maurodatamapper/sde-resources#191 needs to be tested alongside this.

NigelPalmer:
closes #481


giovannaOCC (approved):
All looks good and all tests pass locally too.
It works as it should

PR #498, opened 2024-12-19
gh-497: Fix DataSpec cannot be copied when submitted or attached to request

### LINKED ISSUE #497: DataSpecification can be copied if its status is either 'submitted' or 'attached to request'
# Description

Data Specification cannot be copied if its status is either 'submitted' or 'attached to request'

# Expected behavior

A Data Specification appearing to the users as 'submitted' or 'attached to request' is always 'finalised'. This means that it can be copied for creating a new data specification
Currently is not allowed

giovannaOCC:
closes #497


NigelPalmer (approved):
This is a simple change and solves the issue. Thank you.

Issue #244, opened 2023-02-02
New Request list tests


NigelPalmer:
# Description

Tests are required for the new Request list presentation. (see gh-202) Any existing tests in the files listed below should be reviewed to see if they are still relevant and adapted accordingly. The test files are:

1. src/app/data-explorer/data-class-row/data-class-row.component.spec.ts
2. src/app/data-explorer/data-element-row/data-element-row.component.spec.ts
3. src/app/data-explorer/data-schema-row/data-schema-row.component.spec.ts
4. src/app/shared/data-element-in-request/data-element-in-request.component.spec.ts

PR #507, opened 2025-01-30
Feature/450 org membership requirement for routes

### LINKED ISSUE #450: Prevent a research user from carrying out actions when they are not part of an Organisation
# Description

During registration of new users, there will be a part of the workflow where the research user account is created but they are not yet approved to be part of an Organisation yet. Until they are officially linked to an Organisation, the researcher user is not allowed to carry out a set of actions in `mdm-explorer`.

If a research user is signed in but not part of an organisation...

They will be allowed to access:

- Anonymous access pages e.g. homepage, about, help etc
- The "SDE" page to track their requests
- Their user account page

They will _not_ be allowed to access:

- The "Browse" page
- The "Search" page
- The "Templates" page
- The "Bookmarks" page
- The "Data Specifications" page

# Guards

Angular supports guards on page routes to allow/deny activation of components. This might be a good way to easily prevent page access from happening. See https://angular.dev/guide/routing/common-router-tasks#preventing-unauthorized-access for further details.

# Disable Controls

Apart from guarding page route access, some kind of disabling/hiding of controls will probably be required based on a standard user check.

# SDE Page
The SDE page should not allow the user to create any request types apart from Help requests. They should be able to view the requests that are associated with their user.

OButlerOcc:
Adds a route guard and adjusts controlls based on if the signed in user is a member of an organisation.
Clicking a link wile not being a member redirect s to the request page.

Supported by:

https://github.com/MauroDataMapper/sde-resources/pull/244

https://github.com/MauroDataMapper/sde-core/pull/343

closes #450

### DIFF
--- a/src/app/app.component.html
+++ b/src/app/app.component.html
@@ -24,6 +24,7 @@ SPDX-License-Identifier: Apache-2.0
       [signInLink]="signInLink"
       [signedInUser]="signedInUser"
       [signedInUserProfileImageSrc]="signedInUserProfileImageSrc"
+      [organisationMember]="organisationMember"
       [rightLinks]="headerRightLinks"
       [accountLink]="accountLink"
       [draftDataSpecificationCount]="draftDataSpecificationsCount"
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -255,10 +255,10 @@ export class AppComponent implements OnInit, OnDestroy {
     this.broadcast
       .sdeOnUserOrganisationStatusUpdated()
       .pipe(takeUntil(this.unsubscribe$))
-      .subscribe(() => this.setControlValues(this.userDetails.sdeGetUserOrganisationMembership()));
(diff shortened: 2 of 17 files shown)


NigelPalmer (changes requested):
The Requests menu option has disappeared when you log in, even if you are a member of an organisation.
image

OButlerOcc:
> The Requests menu option has disappeared when you log in, even if you are a member of an organisation. image

This was a symptom of a screw up with the flags, Good thing you caught it before the demo.
Modified the flags to actually work correctly.
Corrected a typo checking for the wrong saved variable.
Added flag checks and saves so it now works when the page is reloaded and sign out correctly clears.

mhocc (approved):
This all appears to be working as required, will probably need `sde-resources#237` merging first to make the build errors go away.

NigelPalmer (approved):

PR #509, opened 2025-02-05
gh-508: Attach Data Spec to Project Enquiry Request

### LINKED ISSUE #508: Project Enquiry Request - Add Data Specification
Support for maurodatamapper/sde-resources#186

NigelPalmer:
close #508


giovannaOCC (approved):
All works as intended and tests are all passing on my local branch.

just one thing, outside the requirements that we might want to think.....
when you attach the data specification to an existing request, id does as it should, BUT if I change my mind and I don't want it anymore, I cannot remove it. I can eventually replace with another one but it doesn't allow me to take it off.

PR #514, opened 2025-02-10
gh-512: Reorganise tabs and menus

### LINKED ISSUE #512: User Portal - Reorganise tabs and menu options
OUH have requested some reorganising of the menu options in the portals.

### SDE User Portal
The menu items should be the following:
Home Organisation Projects Catalogue Requests About Help 

- Home: When clicked this will take the user to the landing page. (This is the same user journey as clicking the Thames Valley & Surrey image in the top right) ✔  
- Organisation: Clicking this will show the contents of the current Requests -> Organisation sub tab. This will no longer be displayed in a sub tab and as it's own page ✔
- Projects:  Clicking this will show the contents of the current Requests -> Projects sub tab. This will no longer be displayed in a sub tab and as it's own page. Raza is working on a task to change the contents of this tab. 
- Catalogue: Turn this into a sub menu (Like the help button - But make sure that when clicked the submenu is always displayed unlike the help button that takes you to a help page unless the arrow is clicked.) The options on the sub menu should be "Browse" and "Search". Clicking "Browse" will have the same affect as the current "Browse" button. Clicking "Search" will have the same affect as the current "Search" button. ✔
- Requests:  Clicking this will show the contents of the current Requests -> Requests sub tab. This will no longer be displayed in a sub tab and as it's own page. ✔
- About: This will work as it does now. It's just the positioning that has changed. ✔
- Help: This will work as it does now. ✔

## Conversation

James Welch
I wonder if we could split it into two sections if that's easy - 'Requests', and 'X', where 'X' is a name we choose for Projects / Departments

Kinga Varnai
I like that
An easy place to see all your requests
And then the collective noun for project/org/dept

James Welch
Understanding, of course, that you can also get to your requests via the project / dept. you made them about / for

Kinga Varnai
Do we think it’s too many tabs to have requests, orgs and projects all separately in fact?
And browse/search could be grouped under Catalogue, and have the internal tab solution that we currently have under requests
So we have About Organisations Projects Catalogue Requests Help

James Welch
Makes sense to me :thumbs_up:
Maybe 'About' should be last
The thinking worked :grinning:

Giovanna
if I may....

I agree with Browse and Search going under "Catalogue" (I was thinking the same)
yes, About shouldn't be the first (not sure if the last though)
still feel there are too many elements in the menu (but I don't have a better suggestion for now)
Finally, when a researchers log into the User Portal they ended up to a "dashboard" page, which is quite nice but once you start to navigate, there is no way to get back there unless you click on the top-left logo, which is not very intuitive. Perhaps a "home" ?? still not happy with my naming.

James Welch
Yes, I agree that some sort of 'home' button. I'm ok with 'home' as a name, tbh

## The original brief was this:

Rename the Departments tab to be Organisations. Then replace the list of departments with the new organisation tree control create in: https://github.com/MauroDataMapper/sde-resources/issues/185

When an organisation is clicked on the right page should show all the information about that Organisation.
When a department is clicked on the right page should show all the information about that Department with the name of the Organisation still displayed.

mhocc:
*   Added "Home" menu item
*   "Moved" organisations page to top level menu — no longer available as a tab on the "Requests" page.  Did this by leaving the `Departments` component as-is and simply embedding it in a new `Organisations` component. 
*   Done the same with "Projects" as with organisations
*   The Requests page now only shows the requests and is no longer in a tab group. Have left the requests page in the `SdeMainComponent`—might benefit from a rename.
*   Not stated in the issue but have assumed the "Browse" and "Search" pages need to be removed after being put under the "Catalogue" menu.
*   Added a "Catalogue" tab which is a drop down menu offering "Browse" and "Search".  The menu is dropped down if any of the menu text is clicked on and does not navigate until an item is chosen.  This change also affects the "Help" menu which I think is consistent.

Closes #512

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,7 +1,7 @@
 {
   "editor.formatOnSave": true,
   "[typescript]": {
-    "editor.defaultFormatter": "esbenp.prettier-vscode"
+    "editor.defaultFormatter": "vscode.typescript-language-features"
   },
   "[javascript]": {
     "editor.defaultFormatter": "esbenp.prettier-vscode"
@@ -10,7 +10,7 @@
     "editor.defaultFormatter": "esbenp.prettier-vscode"
   },
   "[html]": {
-    "editor.defaultFormatter": "esbenp.prettier-vscode"
+    "editor.defaultFormatter": "vscode.html-language-features"
   },
   "[json]": {
     "editor.defaultFormatter": "esbenp.prettier-vscode"
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,7 +1,7 @@
 {
   "editor.formatOnSave": true,
   "[typescript]": {
-    "editor.defaultFormatter": "vscode.typescript-language-features"
+    "editor.defaultFormatter": "esbenp.prettier-vscode"
   },
   "[javascript]": {
     "editor.defaultFormatter": "esbenp.prettier-vscode"
@@ -10,7 +10,7 @@
     "editor.defaultFormatter": "esbenp.prettier-vscode"
   },
   "[html]": {
-    "editor.defaultFormatter": "vscode.html-language-features"
+    "editor.defaultFormatter": "esbenp.prettier-vscode"
   },
   "[json]": {
     "editor.defaultFormatter": "esbenp.prettier-vscode"
--- a/src/app/pages/projects/projects.component.html
+++ b/src/app/pages/projects/projects.component.html
@@ -19,15 +19,6 @@ SPDX-License-Identifier: Apache-2.0
 <div class="container">
     <div class="main-row hero">
         <h1>Projects</h1>
-        <mdm-sde-sign-in *ngIf="!signedIn"></mdm-sde-sign-in>
-        <div class="medium-top-spacer flex-container__card">
-            <sde-requests-list [useConfig]="requestsNeedingApprovalListConfig"></sde-requests-list>
-            <div class="help-text mt-2">
-                <i class="bi bi-info-circle-fill info-icon"></i>
-                <span>
-                    A list of all project scoped requests that require approval at this project.
-                </span>
-            </div>
-        </div>
+        <sde-project-page [isReadOnly]="true"></sde-project-page>
     </div>
 </div>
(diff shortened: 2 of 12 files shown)


.vscode/settings.json
```diff
@@ -10,9 +10,9 @@
...
   },
   "[html]": {
-    "editor.defaultFormatter": "esbenp.prettier-vscode"
+    "editor.defaultFormatter": "vscode.html-language-features"
   },
   "[json]": {
     "editor.defaultFormatter": "esbenp.prettier-vscode"
   }
-}
+}
```
NigelPalmer: Please revert these changes, the removal of the prettier references.

src/app/pages/projects/projects.component.html
```diff
@@ -0,0 +1,33 @@
...
+            <sde-requests-list [useConfig]="requestsNeedingApprovalListConfig"></sde-requests-list>
+            <div class="help-text mt-2">
+                <i class="bi bi-info-circle-fill info-icon"></i>
+                <span>
+                    A list of all project scoped requests that require approval at this project.
+                </span>
+            </div>
+        </div>
+    </div>
+</div>
```
NigelPalmer, line 24: Please get the latest changes from develop and add them to this branch, then update this page to point at the new Projects page. e.g:
image

NigelPalmer (changes requested):
Looking good and working as specified. Reopening so that the new projects page can be hooked up.

.vscode/settings.json
mhocc: Done

src/app/pages/projects/projects.component.html
mhocc: Have rebased onto develop but I don't see a `sde-project-page` anywhere

giovannaOCC: @mhocc  that is because sde-project-page is defined in sde-resources. you basically need to put back the this commit:
#516 
literally you just need to do exactly what suggested, because the component is already in the list of the imported from sde-resources

NigelPalmer: This has now been done and is working as expected.

NigelPalmer (approved):
Everything now looks good.

PR #518, opened 2025-02-14
gh-517: Restore hide menu items

### LINKED ISSUE #517: All menu items when user not part of an organisation
Work was done to hide menu items when a user is not a member of an organisation, but these were lost in a rebase. They need to be restored.

NigelPalmer:
closes #517


OButlerOcc (approved):
Tested fine. Has the changes we made to flag setting on load that stop the flag falling out of sync.

/projects wasn't secured properly, it hid the link but you could put /projects in the url and get there, added it to the route level security. Pushed that change so you need to be an org member and logged in to see /projects.

PR #524, opened 2025-02-28
gh-467: Branding follow-up

### LINKED ISSUE #467: Branding follow-up
A few outstanding branding issues to be mopped up but are not required to block the branding release.

The icon needs to change in the tab on the browsers.

When you have no bookmarks saved, by clicking on the Bookmarks menu on the top-right, you will get a message (alert) saying that you don't have any.
Message is correct but the background is light blue -> it comes from " .alert-info in _alert.scss file (directly from bootstrap)

mhocc:
- Change background for "no bookmarks" to match other panels
- Created a naff new icon for the explorer interface

Closes #467


SystemCRaza (approved):
Approved.

PR #533, opened 2025-03-26
gh-532: removed impersonate endpoint dependencies

### LINKED ISSUE #532: Remove Impersonate endpoint dependencies
An endpoint was removed from sde-core. this is a follow-up task.

SystemCRaza:
Please test this with sde-resources issue 333 and sde-core issue 331 and sde-admin-ui issue 279.

An endpoint was removed from sde-core. this is a follow-up task.


giovannaOCC (approved):
all working as intended

PR #536, opened 2025-04-29
Cleaned query builder css: colour, spacing, padding, margins, border radius, box shadow Fixed bug with date input Increased with of content area on desktop Fixed black font on dark blue, and added grey background to highlighted elements on date calendar

edwardcrichton:


joe-crawford:
Thanks @edwardcrichton! Changes look good :+1:

Commenting that when you select a day in the date picker, that day then disappears from the other months!

edwardcrichton:
@joe-crawford Added some more css to colour in the calendar days so that they can be seen.

joe-crawford (approved):
Thanks Ed, looks like an improvement to me! :+1:

PR #537, opened 2025-04-29
Added "Frutiger", Arial to css Ensured it is used by body and .mat-typography Added @font pointing to https://thamesvalleyandsurreyhealthandcaredata.nhs.uk

edwardcrichton:


joe-crawford:
Tested this branch but it did not work for me sorry! Can see the change in the compiled CSS file but Chrome and Safari are still showing Roboto.

BTW think it should be OK to merge for now, at some later point we should probably look at de-branding this repository and moving the non open source assets etc. somewhere else.

joe-crawford:
Realised I did not update the Mauro API Property, will try again!

edwardcrichton:
@joe-crawford I realised why it might not be working for you as I was heading back on the bus: the mauro api property was being set on my set-up but not yours. So I've set the appropriate css variable and made it !important. Is it working for you now?

joe-crawford:
Thanks Ed, it picked up the changes and most of the fonts show as Arial now on localhost, which makes sense. However I can see these ones are still showing as Roboto/Helvetica/Segoe UI... think these may need some more tweaks to make it consistent?

Screenshot 2025-04-30 at 10 48 37 AM
Screenshot 2025-04-30 at 10 48 43 AM
Screenshot 2025-04-30 at 10 48 47 AM
All on http://localhost:4201/#/dashboard

edwardcrichton:
@joe-crawford Yep, those three were using directly specified fonts, or the material default font list. Is that better now?

joe-crawford:
Thanks @edwardcrichton, yes can only see Arial now, great :+1:

joe-crawford (approved):

PR #538, opened 2025-04-30
Move query text to bottom of query builder, inside a show/hide toggle

edwardcrichton:


joe-crawford:
Thanks @edwardcrichton! Rebased on your other changes. Looks good! Probably we should use the grey elsewhere as well instead of everything being blue!

joe-crawford (approved):

PR #542, opened 2025-05-01
Ability to add data elements to an existing model from the browse component

edwardcrichton:


edwardcrichton:
@joe-crawford A test is failing... I'll figure out why that would be ...

joe-crawford:
Looks good :+1:

Could you have a look at adding the spinner when the data class is added? Especially as the Grails backend can be a bit slow there!

joe-crawford (approved):
Looks great, thank you! 💫

PR #551, opened 2025-05-07
Nicer table borders

edwardcrichton:


joe-crawford (approved):
Definitely an improvement :+1:

PR #554, opened 2025-06-11
More tweaks to work with mauro-micronaut

joe-crawford:
Remove unneeded path lookups. Looks like the DataElement IDs were already present so these weren't needed anyway.


jamesrwelch (approved):
If it works, I like it 👍

